PR #1, opened 2015-07-30
E0d/locust scaffolding

e0d:
@nedbat @sarina

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,4 @@
 # edx-load-tests
 Open source load tests for edX software components
+
+This project is a single source for public load test for any software component.  New test should be developed here unless there is a very good reason why that cannot be the case.  Old test should be scrubbed and moved here over time.
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
 # edx-load-tests
 Open source load tests for edX software components
 
-This project is a single source for public load test for any software component.  New test should be developed here unless there is a very good reason why that cannot be the case.  Old test should be scrubbed and moved here over time.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -1,2 +1,4 @@
 # edx-load-tests
 Open source load tests for edX software components
+
+This project is the single source for public load tests for edX software component.  New tests should be developed here unless there is a very good reason why that cannot be the case.  Old tests should be scrubbed and moved here over time.
```
sarina: I'm thinking this would be a good place to also add something about how the tests are to be used, what sort of support - if any - edX offers, etc.

README.md
```diff
@@ -1,2 +1,4 @@
 # edx-load-tests
```
nedbat: Can we stick with README.rst instead?

README.md
e0d: Agree, but we'll need some tests before this makes too much sense. OK to ask the test writers to add this?

e0d:
@alawibaba @dsjen @BenjiLee this is the new repo

README.md
sarina: Yeah that sounds good.

benpatterson:
@jzoldak fyi

doctoryes:
@e0d :+1:  - when will you merge this?

jzoldak:
[This internal wiki page](https://openedx.atlassian.net/wiki/display/EdxOps/How+to+Run+Performance+Tests) should be updated to reference the new repo and when to use which one.

e0d:
@jzoldak updated the wiki.  It would be nice to have an open version of that wiki page.

PR #3, opened 2015-08-18
New CSM load test structure. Pulls in all of the necessary

alawibaba:
requirements to directly load test the Django ORM XBlockUserStateClient.

PR #4, opened 2015-08-19
Csm load model

cpennington:
@alawibaba @doctoryes @macdiesel: This builds on https://github.com/edx/edx-load-tests/pull/3 to add a model of current CSM load.

Todo:
- [x] Set the weights for get_many and set_many
- [x] Call set_many if get_many can't return enough documents

[PLAT-601]


macdiesel:
LGTM :+1:

doctoryes:
:+1:

PR #6, opened 2015-08-20
Add travis, use pep8.

benpatterson:
This puts basic structure in place for travis integration. With this commit,
it enforces pep8 linting standards.


benpatterson:
@alawibaba @cpennington @doctoryes 

The requirements layout is preventing us from running travis builds. 

At this point, there are two issues:
- dev-requirements needs to be empty in order for requirements to (almost) successfully install. That's cruft so we need to sort out what kind of requirements.txt paradigm we want.
- The string being used for the github requirements is not supported by Travis, as per the last failing build.

I'd like to get travis going on this repo, even if it's just pep8 for starters. Could one of you assist?

cpennington:
I'm going to put this on hold until we clean up the requirements formats a little bit, so that pep8 checking doesn't end up installing all of the test-specific requirements (like scipy)

benpatterson:
Solid, thanks @cpennington 

Glad to get the travis starter-kit going. :fireworks:

PR #7, opened 2015-08-20
Teams discussion tests.

dianakhuang:
New load tests to test teams discussion interactions.

Specifically designed to tax the behavior of setting last_activity_at.

Reviewers: @efischer19 and (@benpatterson or @jzoldak )

### DIFF
--- a/locust/teams_discussion/locustfile.py
+++ b/locust/teams_discussion/locustfile.py
@@ -1,0 +1,196 @@
+"""
+Load test to test the performance of teams discussions.
+
+Usage:
+
+  $ locust --host="http://localhost:18080"
+
+Supported Environment Variables:
+
+  TEAM_COURSE_ID - course id that we're running these tests against
+
+"""
+import os
+import sys
+import string
+import json
+import random
+from collections import deque
+
+from locust import HttpLocust, task
+# Work around the fact that this code doesn't live in a proper Python package.
+sys.path.append(os.path.join(os.path.dirname(__file__), '..', 'helpers'))  # noqa
+from auto_auth_tasks import AutoAuthTasks
+from opaque_keys.edx.keys import CourseKey
+
+
+_dummy_chars = string.lowercase + ' '
+
+
+def _dummy_text(minlen, maxlen):
+    """
+    Generate dummy text for forum submissions.
+    """
+    return "".join(random.choice(_dummy_chars) for _ in xrange(minlen, random.randrange(minlen + 1, maxlen)))
+
+
+class TeamsDiscussionTasks(AutoAuthTasks):
+    """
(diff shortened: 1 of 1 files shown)


locust/teams_discussion/locustfile.py
```diff
@@ -0,0 +1,175 @@
...
+from collections import deque
+
+from locust import HttpLocust, task
+# Work around the fact that this code doesn't live in a proper Python package.
+sys.path.append(os.path.join(os.path.dirname(__file__), '..', 'helpers'))
+from auto_auth_tasks import AutoAuthTasks
+from opaque_keys.edx.keys import CourseKey
+
+
+TEAMS_FILE = 'teams.json'
```
benpatterson: Does this need to be in an `os.getenv`?

dianakhuang: I do that down in line 53. Is it too confusing this way?

benpatterson: I guess what I mean is, if someone wants to name their file something else like `'test1.json` and use that as the env var. I was partly raising it for consistency with TEAM_COURSE_ID, and also partly to avoid touching the script if I have a different json file.

benpatterson: oh oh I see what you mean. Er, I suppose it's a bit confusing. But, not a must.

locust/teams_discussion/locustfile.py
```diff
@@ -0,0 +1,175 @@
...
+        kwargs.update({'headers': self._headers})
+
+        if params is not None:
+            kwargs.update({'params': params})
+
+        return getattr(self.client, method)(path, **kwargs)
+
+    @task(2)
+    def view_team_page(self):
+        """ View the page for the users' current team. """
```
benpatterson: I'm assuming this kind of post is what the GUI would produce? (As opposed to the GUI producing more noise/heavier requests). Just wanted to doublecheck.

benpatterson:
I had some nits but LGTM.

locust/teams_discussion/locustfile.py
dianakhuang: This is the page where the team's discussions are displayed, so it will have to make calls out to the team api and to the  underlying discussion forums. I don't know if that answers your question, though.

efischer19:
I noticed that we're "writing" threads and comments, but only ever "reading" at the team page level. Do we need to consider reading specific threads or comments as potential actions as well?

dianakhuang:
@efischer19 We can write a test for getting back a single thread but not a single comment. I'll add that now.

efischer19:
LGTM. :+1:, pending any changes we find are needed when running against the sandbox.

PR #12, opened 2015-08-27
 Transaction support to the CSM load test.

alawibaba:


alawibaba:
@cpennington PTAL

csm/locustfile.py
```diff
@@ -184,10 +185,12 @@ def get_many(self):
             )
 
     @task(1)
+    @transaction.commit_manually
     def set_many(self):
         usage_key = self._gen_usage_key()
         self.client.set_many(self.client.username, {usage_key: self._gen_block_data()})
         self.usages_with_data.add(usage_key)
+        transaction.commit()
```
cpennington: In your tests in devstack, did the transaction that had the `set` in it also have `get`s?

cpennington: I would have expected at least one call to `get` to retrieve the data that's being `set`.

cpennington:
:+1:

PR #13, opened 2015-08-27
Metric resolution

cpennington:
@alawibaba @macdiesel @doctoryes: This allows us to configure how many metrics to log, and also separates the file logs based on whether the locusts are hatching, or running.

### DIFF
--- a/helpers/raw_logs.py
+++ b/helpers/raw_logs.py
@@ -5,7 +5,7 @@ from datetime import datetime
 from locust import events
 from tempfile import NamedTemporaryFile
 
-LOG_FILE_NAME = "requests-{testid}-{pid}.log"
+LOG_FILE_NAME = "requests-{testid}-{pid}-{state}.log"
 LOG_FIELDS = [
     "start_time",
     "end_time",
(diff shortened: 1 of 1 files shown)


helpers/raw_logs.py
```diff
@@ -36,23 +44,46 @@ def _open_log_file(self):
...
+    def on_hatch_complete(self, **kwargs):
+        self.hatching = False
+        self._metric_counter = 0
+        self._close_log_file()
+
+    def _skip_entry(self):
+        """
+        Return True if the logger should skip this entry.
+        """
+        self._metric_counter += METRIC_RESOLUTION
```
doctoryes: If the first thing done when calling `_skip_entry` is adding the `METRIC_RESOLUTION`, how can `_metric_counter` ever become negative?

cpennington: It won't become negative. It starts at 0, and gets `METRIC_RESOLUTION` added whenever we try to log a metric. If it gets greater than 1, then we *actually* log a metric, and then subtract 1. So, if `METRIC_RESOLUTION` is `0.3`, then we'd see:

    0.3, 0.6, 0.9, *1.2*, 0.5, 0.8, *1.1*, 0.4, 0.7, *1.0*, 0.3, ...

doctoryes: Ah - I was thinking that `METRIC_RESOLUTION` would be greater than 1. Perhaps adding a comment and/or a limit on the value when reading it above would be helpful.

cpennington: Good call. Added.

doctoryes:
:+1:

PR #17, opened 2015-09-08
Add transactions and Pareto limit.

doctoryes:
@cpennington Transactions plus Pareto distribution cap. Please review.


cpennington:
:+1:

PR #19, opened 2015-09-14
All blocks have three fields to avoid growth over time.

doctoryes:
@cpennington Simple change to review - is this what we'd discussed?


cpennington:
:+1:

PR #20, opened 2015-09-16
Iterate over the results returned by get_many.

doctoryes:
get_many was not actually being called since it is a Python generator.


doctoryes:
@cpennington Please review.

cpennington:
:+1:

PR #24, opened 2015-09-25
Send request response times to DataDog.

doctoryes:
@cpennington Please review. (I've tested this and seems to work.)


cpennington:
I was thinking that you'd add this to edx platform, so that we could track the numbers in prod.

doctoryes:
Good point - closing.

PR #27, opened 2015-10-20
Clee/dapi locust

BenjiLee:
@benpatterson @macdiesel 

Optional: @jimabramson 

This is a new loadtest for the new Discussions API.

Data seeding scripts were removed from this PR. 

FYI @nasthagiri


discussions_api/tasks/dapi_tasks.py
```diff
@@ -0,0 +1,426 @@
...
+        PATCH_thread
+        """
+        data = {"abuse_flagged": random.choice(["true", "false"])}
+        name = "abuse_flag_thread" if self.verbose else "PATCH_thread"
+        thread_id = self._get_thread_id()
+        self.edit_thread(
+            thread_id=thread_id,
+            data=data,
+            name=name,
+
```
BenjiLee: will remove space

discussions_api/dapi.py
```diff
@@ -0,0 +1,376 @@
...
+import os
+import random
+import sys
+import urllib
+
+from tasks import dapi_constants
+
+# Work around the fact that this code doesn't live in a proper Python package.
+sys.path.append(os.path.join(os.path.dirname(__file__), '..', 'helpers'))
+from auto_auth_tasks import AutoAuthTasks
```
BenjiLee: Can't figure out why python won't let me import this :( Copied this work around from another locustfile. Is there a way to quiet pep8?

BenjiLee: Figured it out, requirements file installs itself as a package.

discussions_api/dapi.py
```diff
@@ -0,0 +1,376 @@
...
+
+        url = "/api/discussion/v1/threads/?{}&{}&{}&{}&{}&{}&{}".format(
+            course_id,
+            following,
+            view,  # Indexes have not been added.
+            order_by,
+            order_direction,
+            page,
+            page_size,
+        )
```
macdiesel: Instead of all this string replacement you can use urllib more directly. Just create a dict with your query string pairs and use urllib.urlencode that will create the query string. It's easier to read and you'll also get all of the urlencoding you want with less code.

discussions_api/dapi.py
```diff
@@ -0,0 +1,376 @@
...
+        if response.json()["results"]:
+            if prioritize_comments:
+                threads_with_comments = []
+                for thread in response.json()["results"]:
+                    if thread["comment_count"] > 0:
+                        threads_with_comments.append(thread)
+                if threads_with_comments:
+                    return random.choice(threads_with_comments)
+            return random.choice(response.json()["results"])
+        return None
```
macdiesel: This can be simpler. combine the if statements on line 147 & 148. threads_with_comments and the for loop after can a list comprehension:
threads_with_comments = [thread in response.json()['results'] if thread["comment_count"] > 0]
(or something close to this)

all of this code should just move closer to the left hand margin and you should be able to simplify the return statements as you have 3 of them here. 1 or 2 in an if/else statement is good, this becomes a bit complicated.

discussions_api/dapi.py
```diff
@@ -0,0 +1,376 @@
...
+            name = "comment_count={}_response_count={}".format(comment_count, response_count)
+        else:
+            name = "GET_comment_list"
+
+        response = self.client.get(url, verify=False, name=name)
+
+        self.verify_response(response, url)
+        if response.json()["results"]:
+            return random.choice(response.json()["results"])
+        return None
```
macdiesel: Same as above, you could make better use of urllib in this method as well.

discussions_api/dapi.py
```diff
@@ -0,0 +1,376 @@
...
+
+        # Required for question threads, set to False for discussion
+        if thread["type"] == "question":
+            endorsed = "endorsed=False"  # unless set to false, no comments are returned since endorse is not seeded
+        else:
+            endorsed = "endorsed=None"
+
+        #  These parameters did not seem to affect response times 09/29/16
+        # mark_as_read = "mark_as_read={}".format(random.choice(dapi_constants.MARK_AS_READ))
+        # page = "page={}".format(random.choice(dapi_constants.PAGE))
```
macdiesel: I would argue that even if they do not affect response times now, they should be included in the test as they may affect response times in future iterations of the code.

discussions_api/dapi.py
```diff
@@ -0,0 +1,376 @@
...
+            # mark_as_read,
+            # page
+        )
+
+        # Only be useful when running task in isolation with pre-set comment data
+        if verbose:
+            comment_count, response_count = get_comment_and_response_count(thread["id"])
+            name = "comment_count={}_response_count={}".format(comment_count, response_count)
+        else:
+            name = "GET_comment_list"
```
macdiesel: cool idea.

macdiesel: Though I may call it something other than verbose.

discussions_api/dapi.py
```diff
@@ -0,0 +1,376 @@
...
+
+    def on_start(self):
+        """
+        For Discussion API tests, user creation is done at the parent task level
+        """
+        pass
+
+    def verify_response(self, response, url):
+        """
+        Checks for a 200 response, else prints out the url and
```
macdiesel: Doc could use improvement by adding accepts, returns and most importantly raises.

discussions_api/dapi.py
```diff
@@ -0,0 +1,376 @@
...
+    def __init__(self, *args, **kwargs):  # pylint: disable=super-on-old-class
+        super(DiscussionsApiTasks, self).__init__(*args, **kwargs)
+        self.course_id = os.getenv('COURSE_ID')
+
+        if os.getenv('SEEDED_DATA'):
+            with open("discussions_api/" + os.getenv('SEEDED_DATA'), 'r') as seeded_data:
+                self.thread_id_list = seeded_data.read().splitlines()
+        else:
+            self.thread_id_list = []
+
```
macdiesel: I like that we do not need to have seed data in order to run the tests. :+1:

discussions_api/tasks/dapi_constants.py
```diff
@@ -0,0 +1,49 @@
...
+    "1000char": u"Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Aenean commodo ligula eget dolor. Aenean massa. Cum sociis natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus. Donec quam felis, ultricies nec, pellentesque eu, pretium quis, sem. Nulla consequat massa quis enim. Donec pede justo, fringilla vel, aliquet nec, vulputate eget, arcu. In enim justo, rhoncus ut, imperdiet a, venenatis vitae, justo. Nullam dictum felis eu pede mollis pretium. Integer tincidunt. Cras dapibus. Vivamus elementum semper nisi. Aenean vulputate eleifend tellus. Aenean leo ligula, porttitor eu, consequat vitae, eleifend ac, enim. Aliquam lorem ante, dapibus in, viverra quis, feugiat a, tellus. Phasellus viverra nulla ut metus varius laoreet. Quisque rutrum. Aenean imperdiet. Etiam ultricies nisi vel augue. Curabitur ullamcorper ultricies nisi. Nam eget dui. Etiam rhoncus. Maecenas tempus, tellus eget condimentum rhoncus, sem quam semper libero, sit amet adipiscing sem neque sed ipsum. N",
```
macdiesel: :smiley_cat:

discussions_api/locustfile.py
```diff
@@ -0,0 +1,132 @@
...
+requests.packages.urllib3.disable_warnings()
+
+
+class DiscussionsApiTest(DiscussionsApiTasks):
+    """
+    This is a repeatable baseline test which utilizes the most used requests.
+
+    PATCH, POST, and DELETE can affect the response times of the discussions
+    API over time. Since the GET Thread and GET Thread List are used the most by
+    far, we use GET Thread and GET Thread List as the baseline.
```
benpatterson: I think your baseline should include a certain number of PATCH/POST/DELETES. It might make it harder to understand performance behavior changes if you are adding those in with your tests when you had zero in the first pass. (I'm assuming these kinds of requests are happening today, even if they are at a smaller %.)

discussions_api/locustfile.py
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+
+class FullDiscussionsApiTest(DiscussionsApiTasks):
+    """
+    Runs all the available discussion API endpoints
+
+    This test will use all the available discussion API endpoints. Over time
+    test will change the response time of the course as it POSTs more threads
+    than it DELETEs.
```
benpatterson: Because the underlying store will grow larger, is that what you mean? That sounds like a problematic assumption to me....How will the system scale over time?

benpatterson: [I mean, the assumption that things will just get worse over time....that sounds like a bad assumption. Obviously a growing store is going to be part of the problem space, but not a problem by itself. The reduced response times as we scale...that's a concern.]

discussions_api/tasks/dapi_constants.py
benpatterson: If I had a nickel for every time I saw someone say that in a discussion post.... :dancer:

discussions_api/dapi.py
```diff
@@ -0,0 +1,376 @@
...
+        }
+
+    def stop(self):
+        """
+        Supports usage as nested or top-level task set.
+        """
+        if self.parent != self.locust:
+            self.interrupt()
+
+    def wait_then_stop(self):
```
benpatterson: :+1: on this method...gives a better proxy for users. How long does it wait for?

discussions_api/dapi.py
```diff
@@ -0,0 +1,376 @@
...
+
+        self.verify_response(response, url)
+        if response.json()["results"]:
+            if prioritize_comments:
+                threads_with_comments = []
+                for thread in response.json()["results"]:
+                    if thread["comment_count"] > 0:
+                        threads_with_comments.append(thread)
+                if threads_with_comments:
+                    return random.choice(threads_with_comments)
```
benpatterson: I'm a bit concerned with using random.choice too much. It might be harder to reproduce problems down the line.

discussions_api/dapi.py
```diff
@@ -0,0 +1,376 @@
...
+    def get_random_comment(self, thread, verbose=False):
+        """
+        Common method for GETting a comment list, returns the comment
+
+        Similar to _get_thread, GETting a comment happens significantly more
+        than other HTTP requests so this method will be used to retrieve
+        comment data.
+
+        Args:
+            thread (dict): The thread we are getting hte comment from
```
benpatterson: nit:s/hte/the

discussions_api/dapi.py
benpatterson: Would this be used on an already-filled store?

discussions_api/dapi.py
BenjiLee: It uses locusts own wait function so it uses the same min and max.

discussions_api/dapi.py
BenjiLee: Unfortunately, there needs to be a pool of threads in other to run GET_thread. Since GET_thread is run 3x more than GET_thread_list, I wasn't able to call GET_thread_list for every time GET_thread was called :(

discussions_api/dapi.py
benpatterson: gotcha

discussions_api/locustfile.py
BenjiLee: That's what FullDiscussionsApiTest is for. We decided to pick this test as the baseline since the other requests are called significantly less and the majority of our traffic is through the GET requests. I had PATCH in this test as well, but it was also an insignificant. Although not ideal, not including the DELETEs and POSTs significantly reduces the complexity of these tests and make them repeatable.

benpatterson: Hmm do you expect an increase in POST/PATCH/DELETES with the new functionality?

discussions_api/locustfile.py
BenjiLee: The response time get's slower as we scale. The more threads in a course, the slower the response time :( 

@macdiesel's Results on ramping https://openedx.atlassian.net/wiki/display/PLAT/cs_comments_service+mongoid+4.0.0+upgrade+load+test+results

https://openedx.atlassian.net/wiki/display/MA/Initial+Loadtesting Under the expand "Tests for smaller course which gradually grew"

discussions_api/locustfile.py
BenjiLee: I think what we are expecting from the Mobile usage is 1. hopefully more people to use forums in general, 2. more GET requests since we think the request chunks will be smaller, and 3. more requests against popular threads when notifications is implemented. For POST/PATCH/DELETEs, no significant changes expected.

discussions_api/dapi.py
BenjiLee: In this case, random isn't that random as it is only selecting from a shortened list of threads that are likely to have comments in them. "prioritize_comments" is passed when a GET_comment request needs a thread with comments.

discussions_api/locustfile.py
benpatterson: Ok thanks that's helpful. I think the focus should be apples-to-apples. So, it might make more sense to only do GETs on both baseline + tests. (Or to have the small % of POST/PATCH/DELETE on both baseline + tests).

discussions_api/dapi.py
BenjiLee: Any suggestions?

discussions_api/dapi.py
BenjiLee: Looks like my cat caught on fire while in the middle of  writing this one o.O

discussions_api/tasks/dapi_constants.py
BenjiLee: Do a lot of people meow on the forums? Is our user base a bunch of cats?! Is that why all the pictures on the internet are of cats?!!?!

discussions_api/dapi.py
BenjiLee: Took me a while to figure out that when sending a user back up to the parent task with stop, that it doesn't wait at all, thus the wait then stop.

discussions_api/locustfile.py
BenjiLee: I consider the different tests to be measuring for something different. I would imagine FullDiscussionsApiTest to be a ramping test, GetThreadWithCommentsTest a test to measure pagination of comments, and DiscussionsApiTest a test that measures the common usage.

discussions_api/locustfile.py
macdiesel: I suspect this may have been improved when we added the missing index. That said it will take more time to return responses over the wire as the data size grows though I don't expect to this problem to be quite as extreme as it was during the test scenario that is quoted.

benpatterson: Ok makes sense. End-goal is to qualify what kind of degradation we are looking at, and--keeping with a 200ms standard, at what point we'd need to worry about a caching layer or other kinds of solutions to maintain that standard.

discussions_api/locustfile.py
benpatterson: Ah! Sorry if I missed that nuance. (If it's not documented, you should put that in the doc strings.)

BenjiLee:
Done with first pass. Changes in new commit

discussions_api/locustfile.py
BenjiLee: It's the doc at the top of the locustfile. I can add more details.

BenjiLee:
Ready for another pass.

benpatterson:
:+1:

discussions_api/dapi.py
```diff
@@ -0,0 +1,366 @@
...
+    def verify_response(self, response, url):
+        """
+        Checks for a 200 response, else prints out the url and partial content
+
+        Raises:
+            UnexpectedResponse: raised when we do not get a 200 response.
+        """
+        if response.status_code != 200:
+            print url
+            print response.status_code
```
macdiesel: We should be using a logger instead of print statements.

discussions_api/dapi.py
```diff
@@ -0,0 +1,366 @@
...
+        query_args = {
+            "course_id": urllib.quote_plus(self.course_id),
+            "following": random.choice(dapi_constants.FOLLOWING),
+            "view": random.choice(dapi_constants.VIEW),
+            "page_size": page_size if page_size else random.choice(dapi_constants.PAGE_SIZE),
+            "page": page if page else self.random_page_number(),
+            "order_by": "comment_count" if prioritize_comments else random.choice(dapi_constants.ORDER_BY),
+            "order_direction": "desc" if prioritize_comments else random.choice(dapi_constants.ORDER_DIRECTION),
+        }
+        encoded_args = urllib.urlencode(query_args)
```
macdiesel: :+1:  Thank you

discussions_api/dapi.py
```diff
@@ -0,0 +1,366 @@
...
+        POST a comment (2nd level comment) to the response list
+
+        Args:
+            comment_id (str): the id of the response to create the comment for
+            thread_id (str): The id of the thread to create the comment for
+
+        Returns:
+            (dict): The comment.
+        """
+        url = "/api/discussion/v1/comments/"
```
macdiesel: nit: we should put the url prefix stuff ("/api/discussion/v1") in to it's own variable.

discussions_api/tasks/dapi_tasks.py
```diff
@@ -0,0 +1,424 @@
...
+    @task(1)
+    def abuse_flag_thread(self):
+        """
+        Flag a thread with
+
+        Calls made:
+        GET_thread_list
+        PATCH_thread
+        """
+        data = {"abuse_flagged": random.choice(["true", "false"])}
```
macdiesel: The relative number of threads marked as abuse_flagged in the system are most likely very low. When this code is called it will start flagging ~50% of the threads as abuse_flagged. Could this have an effect on pagination/filtering and the number of threads returned for a user during the tests? I'm not sure this is really representative of the behavior of the users on the site.

discussions_api/dapi.py
```diff
@@ -0,0 +1,366 @@
...
+
+        response = self.client.get(url=url, verify=False, name=name)
+
+        self.verify_response(response, url)
+
+        if not response.json()["results"]:
+            return None
+
+        if prioritize_comments:
+            threads_with_comments = [thread for thread in response.json() if thread["comment_count"] > 0]
```
macdiesel: good stuff.

macdiesel:
few nits but :+1:

discussions_api/tasks/dapi_tasks.py
BenjiLee: It should only be called once per 56 (1+3+51) PATCH requests. Additionally, the PATCH tasks are called 90 times per (40+40+7560+83+92+194+220) times.

macdiesel: ok :+1:

BenjiLee:
@macdiesel Ready for another pass

macdiesel:
:+1:

PR #28, opened 2015-10-27
Add Programs load test

rlucioni:
Includes a task exercising the Programs API's program list endpoint. Also cleans up some aspects of this repo. ECOM-2627.

@jimabramson @cpennington @edx/ecommerce


.travis.yml
```diff
@@ -10,8 +10,8 @@ cache:
   directories:
     - $HOME/.cache/pip
 install:
-    - pip install pip==6.1.0
-    - pip install -r dev-requirements.txt
+    - pip install -U pip
```
clintonb: You also want to install wheel.

rlucioni: Thanks @clintonb.

requirements/base.txt
```diff
@@ -0,0 +1,7 @@
+# Packages required to run load tests
+-e git+https://github.com/edx/locust.git@edx#egg=locustio[scipy]
+-e .
```
rlucioni: @cpennington why is edx-load-tests installed as requirement? Why does this repo even have a setup.py?

.travis.yml
```diff
@@ -10,8 +10,7 @@ cache:
   directories:
     - $HOME/.cache/pip
 install:
-    - pip install pip==6.1.0
-    - pip install -r dev-requirements.txt
+    - pip install -U pip wheel pep8
```
cpennington: can you use the long form of the flag, to make it clearer what it's doing?

README.rst
```diff
@@ -1,28 +1,55 @@
-This project is the single source for public load tests for edX software component.  New tests should be developed here unless there is a very good reason why that cannot be the case.  Old tests should be scrubbed and moved here over time.
+edX Load Tests  |Travis|_
+=========================
+.. |Travis| image:: https://travis-ci.org/edx/edx-load-tests.svg?branch=master
+.. _Travis: https://travis-ci.org/edx/edx-load-tests
```
cpennington: Thanks for doing this cleanup.

requirements/base.txt
cpennington: It's installed as a requirement so that tests can reference anything in the package as importable modules. Locust does some funny things with `sys.path` when it starts a test.

programs/locustfile.py
```diff
@@ -0,0 +1,40 @@
...
+from programs.config import PROGRAMS_API_URL, JWT_AUD, JWT_ISS, JWT_SECRET_KEY
+
+
+class ProgramsTaskSet(TaskSet):
+    """Tasks exercising Programs functionality."""
+    USERNAME_PREFIX = 'load-test-'
+
+    @task
+    def list_programs(self):
+        api = LocustEdxRestApiClient(PROGRAMS_API_URL, session=self.client, jwt=self._get_token())
```
cpennington: Would it make sense to use the techniques here (http://docs.locust.io/en/latest/testing-other-systems.html) to make these tests read more like: `self.client.programs.get()`? That way, the setup of the `LocustEdxRestApiClient` can be handled in common across all programs tests.

helpers/api.py
```diff
@@ -0,0 +1,58 @@
...
+"""
+from edx_rest_api_client import exceptions
+from edx_rest_api_client.client import EdxRestApiClient
+import requests
+import slumber
+
+
+class LocustResource(slumber.Resource):
+    """Custom Slumber Resource which takes advantage of Locust's extended HttpSession."""
+    def _request(self, method, data=None, files=None, params=None):
```
jimabramson: optional: add a **kwargs to the signature, to insulate from errors in case an update of slumber modifies the signature of this non-public method.

programs/README.rst
```diff
@@ -0,0 +1,37 @@
...
+Running
+-------
+
+You can run the Programs load tests from the top level ``edx-load-tests`` directory by executing something like the following:
+
+.. code-block:: bash
+
+    $ JWT_SECRET_KEY=replace-me locust --host=http://localhost:8009/ -f programs
+
+There appears to be a bug in Locust preventing tests from accessing hosts using SSL. To get around this, use ``http`` as the protocol in the host URL, not ``https``.
```
jimabramson: doesn't this seem like a significant issue?  is there an issue logged with locust about it?

programs/README.rst
```diff
@@ -0,0 +1,37 @@
...
+-------------
+
+The load tests rely on configuration which can be specified using environment variables.
+
+==================== ========= =================================
+Variable             Required? Description
+==================== ========= =================================
+PROGRAMS_SERVICE_URL No        URL root for the Programs service
+PROGRAMS_API_URL     No        URL root for the Programs API
+JWT_AUD              No        JWT audience claim (aud)
```
jimabramson: in practice I believe this would have the same value as `PROGRAMS_SERVICE_URL`

programs/README.rst
```diff
@@ -0,0 +1,37 @@
...
+JWT_ISS              No        JWT issuer claim (iss)
+JWT_SECRET_KEY       Yes       Secret key used to sign JWTs
```
jimabramson: in practice this will be the oauth2 client secret key

programs/config.py
```diff
@@ -0,0 +1,25 @@
+"""Configuration for Programs load testing."""
+import os
+
+
+# URL CONFIGURATION
+PROGRAMS_SERVICE_URL = os.environ.get(
+    'PROGRAMS_SERVICE_URL',
+    'http://localhost:8009'
```
jimabramson: i think this is 8004 in devstack.

programs/locustfile.py
```diff
@@ -0,0 +1,40 @@
...
+        api.programs.get()
+
+    def _get_token(self):
+        payload = {
+            'preferred_username': self.USERNAME_PREFIX + str(uuid.uuid4()),
+            'iss': JWT_ISS,
+            'aud': JWT_AUD,
+            'exp': datetime.utcnow(),
```
jimabramson: should the expiration be set to sometime in the future (say now + 24h) just to avoid having to tweak the instance under test from ignoring expiration?

requirements/base.txt
rlucioni: @cpennington it does look like Locust does some `sys.path` manipulation, but it appears to [revert any changes](https://github.com/locustio/locust/blob/master/locust/main.py#L322-L328) it makes afterwards.

Is there a specific import in a test that requires this? The new test I've added doesn't require installing edx-load-tests to import modules from other packages in the repo.

.travis.yml
rlucioni: Updated.

README.rst
rlucioni: No problem.

programs/locustfile.py
rlucioni: @cpennington use of pre-generated JWTs in conjunction with `EdxRestApiClient` currently requires instantiation of the client for each request (to attach the JWT Authorization header). If that weren't the case, we could take the approach described in Locust's docs.

helpers/api.py
rlucioni: @jimabramson instead of silencing that kind of error, wouldn't we prefer this to fail loudly so we can investigate the change to Slumber and modify this accordingly?

jimabramson: @rlucioni actually since this isn't a user-facing system I agree, that makes more sense.

programs/README.rst
rlucioni: @jimabramson it does seem pretty serious. This is actually due to an [issue with gevent](https://github.com/gevent/gevent/issues/477) that appears to have been fixed in 1.0.2. However, Locust is still running the broken 1.0.1.

programs/README.rst
rlucioni: The Programs settings suggest using the value of `SOCIAL_AUTH_EDX_OIDC_KEY` for this. In `config.py`, I've defaulted `JWT_AUD` to the default value used for `SOCIAL_AUTH_EDX_OIDC_KEY`, but I've also made it easily configurable in case it needs to be something else.

programs/README.rst
rlucioni: True, but the Programs settings still require setting a value for `JWT_SECRET_KEY`, a name which I think makes more sense in this context. It should be pretty easy to track down the right value for this setting if you look at the settings on your Programs instance.

programs/config.py
rlucioni: Indeed [it is](https://github.com/edx/configuration/blob/master/vagrant/base/devstack/Vagrantfile#L46), updated.

programs/locustfile.py
rlucioni: Programs uses [DRF-JWT's default](http://getblimp.github.io/django-rest-framework-jwt/#additional-settings) `JWT_EXPIRATION_DELTA`, 5 minutes, which should be more than enough given that requests will be made almost immediately after generating the token.

programs/README.rst
jimabramson: my mistake.  your approach sounds good.

programs/README.rst
jimabramson: ok.

programs/locustfile.py
jimabramson: @rlucioni actually I'm confused by that.   My understanding is that you set `JWT_LEEWAY` on the server side, per the documentation you linked, to permit a small amount of cushion after the stated expiration in the token ([exp](https://tools.ietf.org/html/draft-jones-json-web-token-07#section-4.1)).

It's not very clear from the DRF-JWT doc but it seems like `JWT_EXPIRATION_DELTA` is something that comes into play only when creating JWTs using that library, and it seems to basically be a way of specifying those JWTs' expiration in terms of `now()` plus some lifetime.

programs/locustfile.py
jimabramson: Could `self.client` be reimplemented as an `@property` of a new `TaskSet` subclass or mixin?  It could instantiate and return a `LocustEdxRestApiClient` configured as it is here, which would allow us to basically use that same API.  Too convoluted?

jimabramson:
@rlucioni please note my comment (here)[https://github.com/edx/edx-load-tests/pull/28#discussion_r43166485] which is hidden under an outdated diff.

programs/locustfile.py
cpennington: I thought about that, and it's ok as long as you really want the client to be per-request. That seems a bit odd to me, though, honestly. Shouldn't the client be per-virtual-user?

jimabramson: I'd rather it works that way too, since it seems wasteful to be constantly regenerating clients - esp. when we're trying to maximize throughput in a load test.  It would be great to reuse the client somehow if it doesn't require bending slumber all out of shape.

programs/locustfile.py
rlucioni: @jimabramson my bad, you're right. I will set this to "now + 24 hours."

requirements/base.txt
cpennington: It's possible that the need for this was resolved when I restructured the package. However, I think it was added so that tests could use code in the `utils` package without needing to do their own `sys.path` manipulation.

programs/locustfile.py
```diff
@@ -0,0 +1,40 @@
+from datetime import datetime
+import uuid
+
+import jwt
+from locust import TaskSet, task, HttpLocust
+
+from helpers.api import LocustEdxRestApiClient
+from programs.config import PROGRAMS_API_URL, JWT_AUDIENCE, JWT_ISSUER, JWT_SECRET_KEY
```
cpennington: If you pull in `helpers.raw_logs`, then you can get csv files with all of the requests in them, and use the ipython notebook that's also checked into the repo to get better stats than locust gives on its own.

programs/locustfile.py
rlucioni: @cpennington @jimabramson updated! The client is now instantiated per-user. Since other tests may wish to implement their client differently, I've intentionally limited these changes to the Programs tests.

programs/locustfile.py
rlucioni: @jimabramson following up here: the expiration time is now a configurable number of days, to support long-running soak tests.

programs/locustfile.py
rlucioni: @cpennington it would be great to document what these raw logs provide and how to use them in this repo's README. Will the raw logs take the place of Locust's statistics reporting, or are they in addition?

Given that I'm presently only after a 95% percentile response time for a single endpoint, and that our fork of Locust already includes error bands in the compiled report post-test, I'm leaning towards not pulling in the raw logging as part of this PR. If we need it later, my impression is that we can pull it in easily.

requirements/base.txt
rlucioni: @cpennington I've verified that the tests can use code from `helpers` without any `sys.path` manipulation, without installing `edx-load-tests` as a dependency. In an effort to clean up this repo, I'm going to remove this and `setup.py` in a separate commit. Let me know if you object.

rlucioni:
@jimabramson @cpennington I've addressed all of your comments. Could you please give this another pass?

jimabramson:
:+1:

cpennington:
:+1:

PR #29, opened 2015-10-28
Make teams_discussion tests conform to expected repo layout

rlucioni:
This is cleanup I noticed we needed while working on the Programs load tests.

@cpennington @dianakhuang


dianakhuang:
:+1:

PR #30, opened 2015-10-28
Keep length of usernames within Django's prescribed limits

rlucioni:
ECOM-2627

FYI @edx/ecommerce

### DIFF
--- a/programs/locustfile.py
+++ b/programs/locustfile.py
@@ -25,6 +25,9 @@ class ProgramsUser(HttpLocust):
     user should wait between executing tasks. This class also provides a custom client used
     to interface with edX REST APIs.
     """
+    # Django's AbstractUser, subclassed by Programs, limits usernames to 30 characters.
+    # See: https://github.com/django/django/blob/stable/1.8.x/django/contrib/auth/models.py#L385
+    USERNAME_MAX_LENGTH = 30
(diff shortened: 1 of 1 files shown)


programs/locustfile.py
```diff
@@ -47,8 +50,14 @@ def __init__(self):
         )
 
     def _get_token(self):
+        if len(self.USERNAME_PREFIX) > self.USERNAME_MAX_LENGTH:
+            raise RuntimeError('A JWT secret key is required to run Programs load tests.')
```
peter-fogg: I realize this is already merged, but this code is pretty confusing to me -- it looks like `USERNAME_PREFIX` and `USERNAME_MAX_LENGTH` are intended to be constants, in which case this conditionally is statically known to be false. Beyond that, I'd expect this error message to say something about the length of the username.

rlucioni: @peter-fogg good catch, that message is definitely wrong; I grabbed it from the config module in this package and forgot to update it. That aside, you make a good point about us knowing this conditional to be false given that we're working with constants. I will fix these issues in an upcoming PR.

rlucioni: Corrected in https://github.com/edx/edx-load-tests/pull/31.

PR #31, opened 2015-10-29
Reduce min and max wait times for virtual Programs users

rlucioni:
Reducing the min and max wait times allows fewer greenlets to generate more throughput. Also removes an unnecessary conditional performed during JWT generation.

@peter-fogg, please review. FYI @edx/ecommerce

### DIFF
--- a/programs/locustfile.py
+++ b/programs/locustfile.py
@@ -31,8 +31,8 @@ class ProgramsUser(HttpLocust):
     USERNAME_PREFIX = 'load-test-'
 
     task_set = ProgramsTaskSet
-    min_wait = 3 * 1000
-    max_wait = 5 * 1000
+    min_wait = 30
+    max_wait = 100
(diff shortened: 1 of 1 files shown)


cpennington:
Just as a heads up, the smaller your wait time, the less sensitive your test will be to hiccups in the service.

peter-fogg:
:+1: Thanks for doing this cleanup.

rlucioni:
No problem @peter-fogg, thanks for catching this.

@cpennington can you elaborate? One would think that smaller wait times would result in more data points, and thus increased sensitivity to any discrepancies in performance.

rlucioni:
Discussed offline with @cpennington. Shorter wait times paired with small numbers of virtual users may cause a load test to miss discrepancies in response time since fewer greenlets get hung up waiting on slow requests.

PR #32, opened 2015-12-16
Moving LMS loadtest to the public repo.

feanil:
Also, I had to add a second '-e' requirement to the base.txt for opaque-keys
since that repo's versions are not properly managed right now.

@e0d  @benpatterson @nedbat @jbarciauskas

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -1,5 +1,10 @@
 # Packages required to run load tests
+# Using -e for now because neither of these packages have released code
+# properly.  Once they have done releases with actualy versions we should
+pin these packages correctly.
 -e git+https://github.com/edx/locust.git@edx#egg=locustio[scipy]
+-e git+https://github.com/edx/opaque-keys.git@9f07da1abf699d10bc3252d3081017e8aa37c302#egg=opaque-keys
+
 edx-rest-api-client==1.2.1
 jupyter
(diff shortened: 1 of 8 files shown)


feanil:
I'll fix the PEP* violations when I have some free time but if one of you has bandwidth please feel free to push the fixes onto this branch.

e0d:
:+1: pending pylint

benpatterson:
I've fixed the violations, though, I had to ignore a rule to do so. The violation we're ignoring is:
`lms/lms.py:13:1: E402 module level import not at top of file`. 

It's hacky to begin with, so I don't feel great about disabling it. However, I don't have time to ensure a change to the file will/will-not impact how locust is run, so I'm leaving it. Someone else can feel free to tackle that one or we can leave it as a TODO.

feanil:
@benpatterson @nedbat second thumb?

benpatterson:
Oh. :+1:

requirements/base.txt
```diff
@@ -1,5 +1,10 @@
 # Packages required to run load tests
+# Using -e for now because neither of these packages have released code
+# properly.  Once they have done releases with actualy versions we should
+pin these packages correctly.
```
nedbat: This line should have a comment marker, right?

PR #34, opened 2015-12-17
Moving more load tests from the private repo to the public one.

feanil:
@benpatterson @e0d


e0d:
@benpatterson @feanil 

violation fixed.

benpatterson:
:+1:

PR #37, opened 2016-01-05
Add missing requirement used in lms/lms.py

maxrothman:
The "lazy" module is used all over lms/lms.py but it doesn't seem to be in any requirements files. Not having it causes locust to not start.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -9,3 +9,4 @@ edx-rest-api-client==1.2.1
 jupyter
 runipy
 seaborn
+lazy


benpatterson:
:+1: from me..

maxrothman:
@benpatterson should I get other thumbs, or is yours sufficient?

benpatterson:
Honestly for this kind of change and in this repo, I consider one sufficient.

But a couple of your co-workers have been pretty involved with this repo, and you could tag them if you want additional looks, or even an FYI

benpatterson:
[so, uh, whatever you like...no objections to a merge :) ]

maxrothman:
@edx/devops

jibsheet:
:+1:

PR #38, opened 2016-01-12
Stop using fork of locust to unblock loadtests

maxrothman:
Sometime in the last 30 days, using our fork of locust broke locust's ability to drive significant load. I observed the following behavior:

* Request 200 users at 10 users/s
* Locust begins spawning users at the requested rate, req/s increases linearly with users
* Locust user spawn rate and req/s start decreasing. Req/s levels off to a test-specific constant (varies between tests but not between runs of tests)
* Locust continues to spawn users albeit at a vastly lower rate. Req/s decrease slowly

Switching to the standard locust distribution solves the issue.

FYI @edx/devops @doctoryes @cpennington

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -1,10 +1,10 @@
 # Packages required to run load tests
-# Using -e for now because neither of these packages have released code
+# Using -e for now because none of these packages have released code
 # properly.  Once they have done releases with actualy versions we should
 # pin these packages correctly.
--e git+https://github.com/edx/locust.git@edx#egg=locustio[scipy]
 -e git+https://github.com/edx/opaque-keys.git@9f07da1abf699d10bc3252d3081017e8aa37c302#egg=opaque-keys
 
+locustio==0.7.3
 edx-rest-api-client==1.2.1
 jupyter
 runipy


feanil:
:+1:

jibsheet:
:+1:

benpatterson:
Might want to include @cpennington I believe he originally brought in the fork...

cpennington:
I'd love to have more information about what's happening on the load drivers. Are they running out of memory? Disk space? Errors on spawn?

Our locust fork has a number of improved reporting capabilities, which is why I'm loath to get rid of it.

maxrothman:
@cpennington I was unable to determine what the root cause was. System resources (memory, disk space, CPU, file handles) were not being depleted, and no errors were shown.

cpennington:
What test were you running?

maxrothman:
I ran the standard LMS test, the course enrollment API test (both from the public repo), and [this test](https://gist.github.com/maxrothman/ce746a8ed3752ecba972). All exhibited the same behavior.

cpennington:
DId you look at network saturation. There was something I added to sync configurations across workers (were you using multiple workers?)

doctoryes:
Max,

Look at this recent commit:
https://github.com/locustio/locust/commit/b026baf63e29326dc4250231e245d6d04718d48d

Perhaps it's the gevent version that's the problem?

On Wed, Jan 13, 2016 at 11:06 AM, Calen Pennington <notifications@github.com

maxrothman:
@cpennington sorry, forgot about that. I checked network saturation as well. We were pushing hundreds of KB over the wire. I also ran an independent load-generating Python script I wrote that used threads and it was able to generate significant load, indicating that the issue is locust, not some piece of infrastructure in between locust and the cluster.

@doctoryes that could be the problem. However, this speaks to a larger problem, i.e. that our fork is not currently well-maintained (otherwise we would've kept up-to-date with the mainline changes). I think we should merge this to unblock loadtesting and punt on switching back to the fork until the ownership issue is resolved.

doctoryes:
If this change is merged, it might break some of the load tests that use the added locustio functionality in our fork. Since we don't have any unit tests of our load tests, that breakage won't be immediately apparent. If the repo stays in this state for long enough, someone else will run up against this breakage and waste time hunting it down.

I agree that the edX forks need to be maintained - and I don't want want to block your work. But incorporating that single commit into our locustio fork seems like the best long-term fix here (assuming it fixes the problem).

cpennington:
Another option would be to simply rebase the `edx/edx` branch onto `locustio/master` (or whatever the most recently released locust verison is), and see if that fixes the issue.

maxrothman:
While this change will break some locust tests, it also unbreaks **all** of the locust tests, including the most frequently used ones. I vote for merging this and fixing forward.

clintonb:
I'm not a fan of unmaintained forks. Which tests actually need the forked version of `locustio`? I would much rather install the fork only for those tests. As is, I have been installing `scipy` for the last 5 minutes despite the fact that I have no need for it. That's just silly (even if it is a one-time install).

maxrothman:
@clintonb @cpennington @doctoryes this PR has been sitting for a while, and in the meantime people are building their loadtests off of it. Can I get a dev thumb so I can merge this? If you prefer, I can add a comment to the relevant tests this change breaks and we can fix them at a later date (with one of the aforementioned suggestions).

clintonb:
:+1:

PR #39, opened 2016-01-15
Added ecommerce load tests

clintonb:


clintonb:
FYI @rlucioni

rlucioni:
@clintonb I've pushed a small cleanup commit to this branch and squashed it into yours. I've verified that the tests run with that new commit applied. Merge away!

PR #40, opened 2016-01-21
Added course-discovery load tests

clintonb:


clintonb:
FYI @rlucioni

rlucioni:
@peter-fogg

clintonb:
@peter-fogg did you make any other changes to this branch that have not been committed?

peter-fogg:
@clintonb It looks like you've rebased since I introduced the commit, but I bumped the edx-rest-api-client version [in this commit](https://github.com/edx/edx-load-tests/commit/4cd80d6e47577e0d595d72f535d1dd1c58225f7f).

peter-fogg:
The older version of the API client required authentication in all cases, but these load tests don't need auth. The newer version (`1.4.0`) is cool with not passing auth credentials.

clintonb:
The version change was handled in another commit already on master.

PR #41, opened 2016-01-25
Added basic structure for load testing

tasawernawaz:
ECOM-3483

* Added basic structure for load testing of credential service
* Added load tests for list endpoint of credential service.

### DIFF
--- a/credentials/config.py
+++ b/credentials/config.py
@@ -1,0 +1,21 @@
+"""Configuration for Credential load testing."""
+import os
+
+
+# URL CONFIGURATION
+CREDENTIAL_SERVICE_URL = os.environ.get(
+    'CREDENTIAL_SERVICE_URL',
+    'http://localhost:8150'
+).strip('/')
+
+CREDENTIAL_API_URL = os.environ.get(
+    'CREDENTIAL_API_URL',
+    '{}/api/v1/'.format(CREDENTIAL_SERVICE_URL)
+)
+
+
+# JWT CONFIGURATION
+JWT_AUDIENCE = os.environ.get('JWT_AUDIENCE', 'credentials-key')
+JWT_ISSUER = os.environ.get('JWT_ISSUER', 'http://localhost:8000/oauth2')
+JWT_EXPIRATION_DELTA = int(os.environ.get('JWT_EXPIRATION_DELTA', 1))
+JWT_SECRET_KEY = os.environ.get('JWT_SECRET_KEY', 'credentials-secret')
--- a/credentials/locustfile.py
+++ b/credentials/locustfile.py
@@ -1,0 +1,68 @@
+import datetime
+import uuid
+
+import jwt
+from locust import TaskSet, task, HttpLocust
+from locust.clients import HttpSession
+from locust.exception import LocustError
+
+from helpers.api import LocustEdxRestApiClient
+from credentials.config import CREDENTIAL_API_URL, JWT_AUDIENCE, JWT_ISSUER, JWT_EXPIRATION_DELTA, JWT_SECRET_KEY
+
+
+class CredentialTaskSet(TaskSet):
+    """Tasks exercising Credential functionality."""
+    @task
+    def list_credential_with_username(self):
+        self.client.user_credentials.get(username="user1")
+
+    @task
+    def list_credential_with_username_and_status(self):
+        self.client.user_credentials.get(username="user1", status="awarded")
+
+
+class CredentialUser(HttpLocust):
+    """Representation of an HTTP "user" to be hatched.
+
+    Hatched users will be used to attack the system being load tested. This class
+    defines which TaskSet class should define the user's behavior and how long a simulated
+    user should wait between executing tasks. This class also provides a custom client used
+    to interface with edX REST APIs.
+    """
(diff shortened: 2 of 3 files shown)


tasawernawaz:
@kashifch Please review this.

kashifch:
I think @benpatterson could help with the review

credentials/config.py
```diff
@@ -0,0 +1,24 @@
...
+)
+
+
+# JWT CONFIGURATION
+JWT_AUDIENCE = os.environ.get('JWT_AUDIENCE', 'credentials-key')
+JWT_ISSUER = os.environ.get('JWT_ISSUER', 'http://localhost:8000/oauth2')
+JWT_EXPIRATION_DELTA = int(os.environ.get('JWT_EXPIRATION_DELTA', 1))
+JWT_SECRET_KEY = os.environ.get('JWT_SECRET_KEY', 'credentials-secret')
+
+if not JWT_SECRET_KEY:
```
ahsan-ul-haq: `JWT_SECRET_KEY ` would never be `None` as you are providing default value.

tasawernawaz:
@jimabramson Can you please review this.

jimabramson:
@tasawernawaz it seems like all the JWT-related stuff could be moved to a shared place and reused in both credentials & programs, as well as various other packages, but this is optional.

credentials/locustfile.py
```diff
@@ -0,0 +1,68 @@
...
+
+from helpers.api import LocustEdxRestApiClient
+from credentials.config import CREDENTIAL_API_URL, JWT_AUDIENCE, JWT_ISSUER, JWT_EXPIRATION_DELTA, JWT_SECRET_KEY
+
+
+class CredentialTaskSet(TaskSet):
+    """Tasks exercising Credential functionality."""
+    @task
+    def list_credential_with_username(self):
+        self.client.user_credentials.get(username="user1")
```
jimabramson: In a more realistic test I would expect that user credential fixtures will be installed for a variety of users (and created on the fly).  If, under test, the application is simply accessing the same few rows in the database repeatedly, this will not give a very realistic picture.

tasawernawaz:
@jimabramson At this point I am not sure, mainly because we have different keys for programs as well as for credentials.

credentials/locustfile.py
```diff
@@ -0,0 +1,68 @@
...
+
+    @task
+    def list_credential_with_username_and_status(self):
+        self.client.user_credentials.get(username="user1", status="awarded")
```
jimabramson: the two task methods for this endpoint don't have any relative weights assigned.   Do you expect equal traffic in a production setting? I suggest trying to get a reasonable estimate.

http://docs.locust.io/en/latest/writing-a-locustfile.html#the-weight-attribute

credentials/locustfile.py
tasawernawaz: Yes we have fixture for populating database, and also we have a locust task for creating credentials (task for POST endpoint). 
Can we set orders of tasks in `TaskSet`, if yes then it will really help us in testing different endpoints in this order
* create
* get
* patch
* list

credentials/locustfile.py
```diff
@@ -0,0 +1,68 @@
...
+import jwt
+from locust import TaskSet, task, HttpLocust
+from locust.clients import HttpSession
+from locust.exception import LocustError
+
+from helpers.api import LocustEdxRestApiClient
+from credentials.config import CREDENTIAL_API_URL, JWT_AUDIENCE, JWT_ISSUER, JWT_EXPIRATION_DELTA, JWT_SECRET_KEY
+
+
+class CredentialTaskSet(TaskSet):
```
jimabramson: are you planning to test any other endpoints?   Does the web view certificate need testing, or will those be served from a CDN on the first release?

jimabramson:
@tasawernawaz the basic structure looks on track.  I'm mainly interested in where you intend to take things from here.

credentials/locustfile.py
tasawernawaz: Yes, actually this is a basic structure of load testing, we have follow up tasks to add tasks for other endpoints as well. After this branch get merged we will start adding tasks for other endpoints.

credentials/locustfile.py
tasawernawaz: I think we can't say which endpoint will have more often requests. I will discuss it with team and will add weights.

credentials/locustfile.py
tasawernawaz: Also we have a sub task for adding locust task for credential rendering, which will have most traffic likely.

credentials/locustfile.py
jimabramson: you can put as many requests inside a single task as you like.   (if you do this, the `min_wait` / `max_wait` won't be used in between those requests - it is only inserted between tasks.)

jimabramson: here is one example of some locust tasks where there are multiple requests / conditional logic.  https://github.com/edx/edx-load-tests/blob/master/discussions_api/tasks/dapi_tasks.py  if you look around the repo i think you'll find others too.

tasawernawaz: Yes that was we were thinking previously, but is it right approach to make more requests under single task.

jimabramson:
:+1:

awais786:
:+1:

PR #43, opened 2016-01-28
locust task added for credentials API endpoints

tasawernawaz:
ECOM-3483

Added locust tasks for
* user credential `list` endpoint
* program credential `list` endpoint
* user credential `POST`, `GET` and `PATCH` requests.
* credential rendering

### DIFF
--- a/credentials/locustfile.py
+++ b/credentials/locustfile.py
@@ -1,24 +1,135 @@
+from collections import deque
 import datetime
+import random
 import uuid
 
 import jwt
-from locust import TaskSet, task, HttpLocust
+from locust import task, HttpLocust
 from locust.clients import HttpSession
 from locust.exception import LocustError
 
 from helpers.api import LocustEdxRestApiClient
-from credentials.config import CREDENTIAL_API_URL, JWT_AUDIENCE, JWT_ISSUER, JWT_EXPIRATION_DELTA, JWT_SECRET_KEY
+from credentials.config import CREDENTIAL_API_URL, CREDENTIAL_SERVICE_URL, JWT, LMS_ROOT_URL, PROGRAM_ID
+from helpers.auto_auth_tasks import AutoAuthTasks
 
 
-class CredentialTaskSet(TaskSet):
+class CredentialTaskSet(AutoAuthTasks):
     """Tasks exercising Credential functionality."""
-    @task
-    def list_credential_with_username(self):
-        self.client.user_credentials.get(username="user1")
+
+    # Keep track of credential ids created during a test, so they can be used to formulate read requests, and patch
+    # requests. A deque is used instead of a list in order to enforce maximum length.
+    _user_credentials = deque(maxlen=1000)
+
+    def on_start(self):
+        self.user_credential_client = self.get_credential_client()
+
+    def get_credential_client(self):
+        """ New property added for using LocustEdxRestApiClient.
+        Default locust client will remain same for using auto_auth().
+        """
+        return LocustEdxRestApiClient(
+            CREDENTIAL_API_URL,
+            session=HttpSession(base_url=self.locust.host),
+            jwt=self._get_token()
+        )
+
+    def _get_token(self):
+        payload = {
+            'preferred_username': "load-test-" + uuid.uuid4().hex,
+            'iss': JWT["JWT_ISSUER"],
+            'aud': JWT["JWT_AUDIENCE"],
+            'iat': datetime.datetime.utcnow(),
+            'exp': datetime.datetime.utcnow() + datetime.timedelta(days=JWT["JWT_EXPIRATION_DELTA"]),
+            'administrator': True
+        }
+        return jwt.encode(payload, JWT["JWT_SECRET_KEY"])
 
     @task
-    def list_credential_with_username_and_status(self):
-        self.client.user_credentials.get(username="user1", status="awarded")
+    def list_user_credential_with_username(self):
+        """ Get all credentials for a user."""
+        self.user_credential_client.user_credentials.get(username="user1")
+
+    @task
+    def list_user_credential_with_username_and_status(self):
+        """ Get credentials having awarded status for a user."""
+        self.user_credential_client.user_credentials.get(username="user1", status="awarded")
+
+    @task
+    def list_program_credential_with_programs_id(self):
+        """ Get credentials of any any status for a program id."""
+        self.user_credential_client.program_credentials.get(program_id=PROGRAM_ID)
+
+    @task
+    def list_program_credential_with_program_id_and_status(self):
+        """ Get credentials having awarded status for a program id."""
+        self.user_credential_client.program_credentials.get(program_id=PROGRAM_ID, status="awarded")
+
+    @task
+    def post_credential_with_attribute(self):
+        # We are creating new user on LMS for generating its credentials,
+        # user information will be used in rendering of credentials.
+        self.auto_auth(hostname=LMS_ROOT_URL)
+        data = {
+            "username": self._username,
+            "credential": {"program_id": PROGRAM_ID},
+            "attributes": [{"name": "whitelist_reason", "value": "Reason for whitelisting."}]
+        }
+        user_credential = self.user_credential_client.user_credentials.post(
+            data=data,
+            name="/api/v1/user_credentials/(with_attributes)"
+        )
+        self._user_credentials.append((user_credential["id"], user_credential["uuid"]))
+
+    @task
+    def post_credential_without_attribute(self):
+        # We are creating new user on LMS for generating its credentials,
+        # user information will be used in rendering of credentials.
+        self.auto_auth(hostname=LMS_ROOT_URL)
+        data = {
+            "username": self._username,
+            "credential": {"program_id": PROGRAM_ID},
+            "attributes": []
+        }
+        user_credential = self.user_credential_client.user_credentials.post(
+            data=data,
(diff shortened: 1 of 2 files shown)


tasawernawaz:
@jimabramson I have added more task, please review this.

credentials/locustfile.py
```diff
@@ -1,24 +1,82 @@
...
+from credentials.config import CREDENTIAL_API_URL, JWT_AUDIENCE, JWT_ISSUER, JWT_EXPIRATION_DELTA, JWT_SECRET_KEY, \
+    LMS_ROOT_URL
+from helpers.auto_auth_tasks import AutoAuthTasks
 
 
-class CredentialTaskSet(TaskSet):
+class CredentialTaskSet(AutoAuthTasks):
     """Tasks exercising Credential functionality."""
+
+    @property
```
jimabramson: shouldn't this be instantiated once per task set instance?  in the `user_credential` task, for example, three of these objects are being instantiated but it seems like just one would suffice.

jimabramson:
@tasawernawaz what is the reason for introducing the slumber client?

credentials/locustfile.py
```diff
@@ -1,24 +1,82 @@
...
-    def list_credential_with_username_and_status(self):
-        self.client.user_credentials.get(username="user1", status="awarded")
+    def list_program_credential_with_program_id_and_status(self):
+        self.user_credential_client.program_credentials.get(program_id=3, status="awarded")
+
+    @task
+    def user_credential(self):
+        # We are creating new user on LMS for generating its credentials,
+        # user information will be used in rendering of credentials.
+        self.auto_auth(hostname=LMS_ROOT_URL)
```
jimabramson: why is it necessary to create users in the LMS?

tasawernawaz: I have added a comment about this. We need users in LMS because when we call render credentials it will hit programs service for programs data and LMS for getting user information from user API.

tasawernawaz:
@jimabramson Slumber client is actually used for authentication through JWT keys, Our `EdxRestApiCleint` is using Slumber in backend, and we have use it in programs locust tests as well.

credentials/locustfile.py
tasawernawaz: Previously we have slumber client instantiated in `CredentialUser`, but it was overriding `locust.client`. But we need `locust.client` for `auto_auth`, we can't send request with slumber client to `auto_auth()` in LMS.

credentials/locustfile.py
jimabramson: got it thanks

credentials/locustfile.py
```diff
@@ -1,24 +1,82 @@
...
+        data = {
+            "username": self._username,
+            "credential": {"program_id": 3},
+            "attributes": []
+        }
+        user_credential = self.user_credential_client.user_credentials.post(data=data)
```
jimabramson: if `self.user_credential_client` must be a `@property` then at least store in a variable in this method and reuse it.   i don't see any reason to instantiate that object three times (once for each request for the service)

jimabramson:
@tasawernawaz have an issue with the LocustEdxRestApiClient lifecycle but it is a minor one.

Will you be adding student certificate views to this test?
Any traffic projections?

credentials/locustfile.py
tasawernawaz: I have updated code for using `user_credential_client`, I have moved this to `on_start()` method. Does it make sense ?

tasawernawaz:
@jimabramson Yes I have added a task for credential rendering as well. I am not clear about but I think api list endpoint (which will be called from learner dashboard) and credential rendering have most of the traffic then other endpoints. 

Could you please review it.

credentials/locustfile.py
jimabramson: yes i think that makes more sense, thanks

credentials/locustfile.py
```diff
@@ -1,24 +1,115 @@
...
+        self._credential_ids.append(user_credential["id"])
```
jimabramson: @tasawernawaz if you save a tuple instead of `(id, uuid)` then you won't need to make an extra call to `get_user_credential` on https://github.com/edx/edx-load-tests/pull/43/files#diff-f07520e67217edcf98937ee3be1266c6R107 just to get the uuid back later.

credentials/locustfile.py
```diff
@@ -1,24 +1,115 @@
...
+        # self.user_credential_client.user_credentials(credential_id).get(name="/api/v1/user_credentials/[id]")
+
+    @task
+    def patch_credential(self):
+        if not self._credential_ids:
+            return
+
+        # random credential_id popped from deque because this task may turn it to revoked,
+        # which will not work for rendering
+        credential_id = self._credential_ids.pop(random.choice(self._credential_ids))
```
jimabramson: `deque.pop()` does not accept arguments.  did you mean [`remove()`](https://docs.python.org/2/library/collections.html#collections.deque.remove) ?

also does it really need to be random?   you could just pop() (or popleft()).

credentials/locustfile.py
```diff
@@ -1,24 +1,115 @@
...
+        self._credential_ids.pop()
```
jimabramson: this isn't needed, i don't think

credentials/locustfile.py
```diff
@@ -1,24 +1,115 @@
...
+        statuses = ["awarded", "revoked"]
```
jimabramson: it's probably best if you only pop/remove the id if you are revoking it.

jimabramson: also keep in mind that the number of ids that get popped during the test can't be more than the number of credentials generated - otherwise you'll run out of ids all the time!

credentials/locustfile.py
```diff
@@ -1,24 +1,121 @@
...
+        record from the deque. Otherwise during rendering task it will return 404.
+        """
+        if not self._user_credentials:
+            return
+
+        credential_id = random.choice(self._user_credentials)[0]
+        status = random.choice(["awarded", "revoked"])
+        data = {"status": status}
+        if status == "revoked":
+            self._user_credentials.pop()
```
jimabramson: don't you want to retire the id that was chosen to be revoked on L101 above?  Doing pop() here is likely to remove a different one, and leave the revoked one in the pool.

credentials/locustfile.py
```diff
@@ -1,24 +1,121 @@
...
-    def list_credential_with_username(self):
-        self.client.user_credentials.get(username="user1")
+    def post_credential(self):
+        # We are creating new user on LMS for generating its credentials,
+        # user information will be used in rendering of credentials.
+        # attributes will be added randomly.
+        self.auto_auth(hostname=LMS_ROOT_URL)
+        attributes = random.choice([
+            [{"name": "whitelist_reason", "value": "Reason for whitelisting."}], []
+        ])
```
jimabramson: since we know that adding attributes means additional database queries, i'd use separate tasks.   then you can see what difference it makes.

awais786:
@jimabramson its ready for final review.

credentials/locustfile.py
```diff
@@ -1,24 +1,136 @@
...
 import jwt
-from locust import TaskSet, task, HttpLocust
+from locust import task, HttpLocust
 from locust.clients import HttpSession
 from locust.exception import LocustError
 
 from helpers.api import LocustEdxRestApiClient
-from credentials.config import CREDENTIAL_API_URL, JWT_AUDIENCE, JWT_ISSUER, JWT_EXPIRATION_DELTA, JWT_SECRET_KEY
+from credentials.config import CREDENTIAL_API_URL, JWT_AUDIENCE, JWT_ISSUER, JWT_EXPIRATION_DELTA, JWT_SECRET_KEY, \
+    LMS_ROOT_URL, CREDENTIAL_SERVICE_URL
```
jimabramson: minor: would be good to clean up this import.  consider a multiline with parentheses, e.g.
```
from credentials.config import (
    foo,
    bar,
)
```

Alternatively, collapse all these settings into dicts (e.g. ```JWT['audience'], JWT['issuer']```, etc) so that fewer imports are needed.

jimabramson:
code looks good to me.  the username and program_id being used are still hardcoded to single values - were those going to be updated?  are you concerned that using the same parameters for those throughout the test will skew the results?

tasawernawaz:
@jimabramson 

I am using `program_id` because we have relative Demo Program in programs fixtures. If we use any new ID which is not in programs service then we will get error on credentials rendering. 
In username case I think we can hardcode because we will create every time new user, and we are actually testing endpoint which return credential on the base of given `username`. In think we don't need to change of `username` each time. Total number of credential will give us right estimate that how long it took to get credential by user.

jimabramson:
@tasawernawaz Understood about the program id.  Maybe the value should go in config, so a tester can override that easily?

About username, i'm not sure that I agree.   This test will be creating and reading many thousands of rows in the user_credential table, and all of them will have the same value for username.  So as the test runs, the result from `list_user_credential` will quickly grow to a very unrealistic size, and the query that is run to get that result will be selecting all the rows in the table, which is not going to be a useful test.

tasawernawaz:
@jimabramson I understand you concern. But we are not using same username for creating user credentials, 
I am using `"username": self._username,` where user is one who created by `auto_auth()` in above line.

jimabramson:
@tasawernawaz ah - you are right.  Sorry, I misunderstood your earlier explanation.

OK!   :+1:  once whatever the build issue is has been fixed

PR #44, opened 2016-02-02
ECOM-3530 Default username added in config.py

tasawernawaz:
Username added in config instead of hard coded in locustfile.


zubair-arbi:
lgtm :+1:

PR #45, opened 2016-02-02
ECOM-3530 fix for username length

tasawernawaz:
Fix length of username for credentials load tests.

@awais786 @zubair-arbi @ahsan-ul-haq


awais786:
:+1:

PR #46, opened 2016-02-02
Switch License from AGPL to Apache

cgoldberg:
### Purpose:
AGPLv3 is not an appropriate license for the code in `edx-load-tests` repo.  This PR makes the necessary changes to relicense the repo's code from AGPLv3 to Apache v2.

---

### Summary:
The Affero GPL is designed to close the "application service provider loophole" that is not protected by the standard GPL license.  This loophole allows ASPs to use GPL code without distributing their changes back to the open source community.  This additional protection is only relevant when licensing hosted software that users interact with over a network (i.e. SaaS applications).

Since the code in `edx/edx-load-tests` is distributed in standard fashion (not SaaS), it does not make sense to use AGPL.  Since there is no ASP loophole to protect, this is not relevant to the code base.  It is essentially offering the same terms as the standard GPL.  This is confusing, and it is also not inline with edX's open source licensing policy (which uses Apache License rather than GPL for all other open source projects).

---

### Changes made to facilitate the license change:

1. update README.rst 'License' section, changing "AGPLv3" to "Apache License, Version 2.0" 

2. create a NOTICE.txt which asserts copyright to edX and declares "Apache License, Version 2.0" as the license in use.

3. update LICENSE.txt, replacing terms and conditions of the 'AGPLv3' license with those of the 'Apache License, Version 2.0'.

---

/cc @rlucioni

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,662 +1,177 @@
-                    GNU AFFERO GENERAL PUBLIC LICENSE
-                       Version 3, 19 November 2007
 
- Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
- Everyone is permitted to copy and distribute verbatim copies
- of this license document, but changing it is not allowed.
-
-                            Preamble
-
-  The GNU Affero General Public License is a free, copyleft license for
-software and other kinds of works, specifically designed to ensure
-cooperation with the community in the case of network server software.
-
-  The licenses for most software and other practical works are designed
-to take away your freedom to share and change the works.  By contrast,
-our General Public Licenses are intended to guarantee your freedom to
-share and change all versions of a program--to make sure it remains free
-software for all its users.
-
-  When we speak of free software, we are referring to freedom, not
-price.  Our General Public Licenses are designed to make sure that you
-have the freedom to distribute copies of free software (and charge for
-them if you wish), that you receive source code or can get it if you
-want it, that you can change the software or use pieces of it in new
-free programs, and that you know you can do these things.
-
-  Developers that use our General Public Licenses protect your rights
-with two steps: (1) assert copyright on the software, and (2) offer
-you this License which gives you legal permission to copy, distribute
-and/or modify the software.
-
-  A secondary benefit of defending all users' freedom is that
-improvements made in alternate versions of the program, if they
-receive widespread use, become available for other developers to
-incorporate.  Many developers of free software are heartened and
-encouraged by the resulting cooperation.  However, in the case of
-software used on network servers, this result may fail to come about.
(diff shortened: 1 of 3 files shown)


edx-webhook:
Thanks for the pull request, @cgoldberg! I've created [OSPR-1107](https://openedx.atlassian.net/browse/OSPR-1107) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

Feel free to add as much of the following information to the ticket:
- supporting documentation
- edx-code email threads
- timeline information ("this must be merged by XX date", and why that is)
- partner information ("this is a course on edx.org")
- any other information that can help Product understand the context for the PR

All technical communication about the code itself will still be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).



We can't start reviewing your pull request until you've submitted a     [signed contributor agreement](http://open.edx.org/sites/default/files/wysiwyg/individual-contributor-agreement.pdf)     or indicated your institutional affiliation.     If you like, you can add yourself to the [AUTHORS](https://github.com/edx/edx-load-tests/blob/master/AUTHORS)      file for this repo, though that isn't required.     Please see the [CONTRIBUTING](https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.rst)     file for more information.

cgoldberg:
> We can't start reviewing your pull request until you've submitted a signed contributor agreement

I don't have a printer, so this might take me a few days to submit.

quick rant:
A *very* nice enhancement to the edX contribution process would be an electronic means of signing the contributor agreement.  Most open source projects support something like this.

Currently, you provide a document (in MS word format?!) that needs to be printed, then snail mailed (or scanned/emailed).  That is a very onerous process when all the contributor wants to do is *give* you an enhancement to your code.  If electronic submission is not feasible, you could (at the very least) provide the form in PDF format, so the user does not need to use a proprietary program like MS Word or jump through doc conversion hoops just to print and sign it.

I'd be curious if you have ever tracked metrics on how many pull requests never get merged due to contributors not returning a signed agreement.  I'd bet it is > 0% :)

nedbat:
@cgoldberg yeah, we're looking into electronic submission of the agreement.  I don't blame you for the rant, and welcome back :)

nedbat:
@cgoldberg btw, since none of this pull request is actually authored by you, they are just boilerplate from others, we won't need a signed agreement for this pull request.

rlucioni:
Thanks for the contribution @cgoldberg!

cgoldberg:
awesome.. thanks guys! :rocket:

PR #47, opened 2016-02-03
ECOM-3619 weight added in locust tasks

tasawernawaz:
I have added `weight=10` for 
* list endpoint which will be called from learner dashboard.
* credential rendering.

@awais786 @zubair-arbi @ahsan-ul-haq

### DIFF
--- a/credentials/locustfile.py
+++ b/credentials/locustfile.py
@@ -49,7 +49,7 @@ class CredentialTaskSet(AutoAuthTasks):
         }
         return jwt.encode(payload, JWT["JWT_SECRET_KEY"])
 
-    @task
+    @task(1000)
     def list_user_credential_with_username(self):
         """ Get all credentials for a user."""
         self.user_credential_client.user_credentials.get(username=USERNAME)
@@ -126,7 +126,7 @@ class CredentialTaskSet(AutoAuthTasks):
             data=data, name="/api/v1/user_credentials/[id]"
         )
 
-    @task
+    @task(10)
     def render_credential(self):
         """ Render a user credential. """
         if not self._user_credentials:


tasawernawaz:
@clintonb Does this makes sense to add weights for list endpoint and credential rendering. I believe both of these endpoints will hit more frequently.

clintonb:
Let's walk through all of the scenarios:

1. Create/issue a credential after a user completes a course.
2. Re-issue a credential after a "re-run" of credentials.
3. Get a list of all credentials to populate the student dashboard.
4. View a single credential after following a link from the dashboard, email, etc.

We will ignore the use case of creating a certificate because it doesn't occur enough to matter and you have no tests for it anyway.

We issue far fewer credentials than we have students. Of the credentials we issue, the likelihood of updating a credential is somewhat low. Thus, those actions have the lowest weight (1).

Without looking at any real-world data, I would guess the student dashboard is loaded more often than a credential. How much more often depends on the data, which you should pull from New Relic. That will determine more adequate weights for those calls. 10x (compared to the create credential call) seems very low, and might result in an incorrectly high throughput.

Also, note that if rendering the credential also retrieves the data from the DB, you don't need to increase the weight of `get_credential`. That API endpoint would not be used in a normal usage scenario, as the data would be rendered server-side.

tasawernawaz:
@clintonb We have fixtures in both programs service and credential service for setup. 
* Programs fixture will add a Demo Program that contains edX Demo Course. 
* Credential fixture create certificate configuration in credential service which will be assigned to users returned by auto_auth().

You are right about adding more weight on List endpoint of API.

tasawernawaz:
@clintonb I have looked into New Relic. We have 200rpm for learner dashboard and 2rpm for viewing web certificate, so I think its is 1:100 ratio for rendering and listing endpoint.

clintonb:
Sounds good to me. :+1: once those weights are added here.

tasawernawaz:
@clintonb Just want to confirm that should I add 10 to render and 1000 to list endpoint or 1 to render and 100 to list.

cpennington:
@tasawernawaz: What do you think is more representative of the load that the service will be under.

tasawernawaz:
@cpennington List endpoint is our main focus while testing load because it will be called every time dashboard will be called. Since ratio between rendering and list endpoint is 1:100, I am adding 10 to rendering and 1000 to list endpoint.

PR #48, opened 2016-02-08
locust 0.7.2

fredsmith:


pwnage101:
can this be closed? it does not even bump locust, despite the title.

PR #50, opened 2016-03-01
Removed hardcoded username in config

tasawernawaz:
Removed hardcoded username in config, now we can pass it through command line parameters.


awais786:
:+1:

waheedahmed:
:+1:

PR #53, opened 2016-06-06
Upgrade numpy to it's latest version

muzaffaryousaf:
We need to upgrade this to keep the numpy versions in-sync across different repos. 
Platform PR: https://github.com/edx/edx-platform/pull/12444


muzaffaryousaf:
@nedbat @andy-armstrong Would you mind to give it a look.

muzaffaryousaf:
@andy-armstrong @nedbat Different version in this repo would't complaint about VersionConflict, Do you think we should update the version here ?

andy-armstrong:
@muzaffaryousaf I'm not sure. @tasawernawaz can you advise.

muzaffaryousaf:
We don't need to upgrade this one either.

PR #55, opened 2016-06-27
Update Discussion API Load Tests

robrap:
#This branch is mostly based on Wajeeha's branch jia/dapi_MA-2476, which was used for the last round of performance tests, so this is our de facto performance tests.  However, there are commits from this branch and from Chris Lee's branch clee/dapi (upon which this is based), that never made it to master.

This branch also adds a commit that I required to fix some references that allowed this to run for me locally like some of the other tests, and includes some initial docs for how this was run.

I need help determining the state of these commits and what we want.

Reviewers:
- [x] @BenjiLee 
- [x] @wajeeha-khalid 

Post-review
- [x] squash commits

PR #56, opened 2016-07-06
TNL-4937: Set up indexes for load testing.

robrap:
This can be used to set up indexes for load testing.  

I'm punting on the larger question of getting migrations set up in the comment service because we are considering a move to mysql.  Once this gets settled, we can determine whether this effort is warranted.

For now, this script is being added to the edx-load-tests repo so we can make progress on the performance testing that will help us make appropriate decisions.

FYI:
@ormsbee 
@dianakhuang 
@jibsheet 
@e0d


robrap:
@benpatterson @jzoldak: FYI: I set up this new PR and the single check was never set up.

robrap:
FYI: @BenjiLee

jzoldak:
@robrap what do you mean by the single check was never set up?

robrap:
@jzoldak: here is another pr for this repo with Travis CI check: https://github.com/edx/edx-load-tests/pull/55

I'll amend and push to see if it fixes.

jzoldak:
@robrap The travis config file for this repo is set up to only run tests for PRs that are against master. Since this is against your other branch, you would need to add that to the config. See [.travis.yml#L18](../blob/master/.travis.yml#L18).

robrap:
@jzoldak:  Cool.  Thank you for explaining.

robrap:
FYI: @tobz @pwnage101

PR #57, opened 2016-07-08
Change the default course_id

pwnage101:
This change allows ansible-provision-locust-master to work out of the box.

### DIFF
--- a/lms/lms.py
+++ b/lms/lms.py
@@ -25,7 +25,7 @@ class EdxAppTasks(AutoAuthTasks):
         """
         The complete id of the course we're configured to test with.
         """
-        return os.getenv('COURSE_ID', 'edX/DemoX/Demo_Course')
+        return os.getenv('COURSE_ID', 'course-v1:edX+DemoX+Demo_Course')
 
     @lazy
     def course_key(self):


jibsheet:
Do we need to update all the hardcoded problem blocks in the rest of the tests?
I assume this is for running against sandboxes vs the loadtest env (which still uses that old draft module and I would love for it to use split, but lots of other things need updating)

pwnage101:
Quoting Kevin Falcone (2016-07-08 16:25:06)
> Do we need to update all the hardcoded problem blocks in the rest of the tests?

It's possible, this only addresses the lms tests.  The other tests are broken for
various different reasons:

https://openedx.atlassian.net/wiki/display/ENG/State+of+Locust

> I assume this is for running against sandboxes vs the loadtest env (which still
> uses that old draft module and I would love for it to use split, but lots of
> other things need updating)

Correct.  Ah, I wasn't thinking of the loadtest env.  In that case I'd rather
not merge this because it would make it impossible to run tests against the
loadtest env since evironment variables are not currently configurable (unless
you shell in, set COURSE_ID, then re-run locust).

Troy

jibsheet:
I was referring to the lms tests, they have hardcoded blocks in there, I can find them on Monday if you're curious.

I would love the loadtest env to also use a split course (not a draft course) by default, I currently shell in, change the course ID and modify tests to use other blocks.  It's a PITA and not complete by any measure.

PR #58, opened 2016-07-21
implement settings interface

pwnage101:
This implements the helpers.settings module which abstracts all settings file
discovery and loading.  Settings files live under the top level
"settings_files" directory in yaml files.  The names of the settings files
correspond to the name of the test module.  E.g. ecommerce test settings are in
ecommerce.yml.

setup.py is introduced to help the settings module discover the settings files
(via the package_data paramter).

Additionally, all `__init__.py` files and relevant module imports are cleaned up
so that every test is invoked the same way (from the root of the repo, the way
that is described in README.md).

Note: this commit completely moves away from using environment variables for
test configuration.

Addresses PERF-352

Testing Done
========
So far, I've tested the following load tests:
* course_discovery
* course_import
* discussions_api
* enrollment
* lms
* csm
* teams_discussion

I skipped these for various reasons:
* credentials - generating JWTs (PERF-361)
* ecommerce - generating JWTs (PERF-361)
* programs - service being retired


pwnage101:
@cpennington @tobz @maxrothman

.pep8
```diff
@@ -2,4 +2,4 @@
 # E501: Line too long.
 # See: http://pep8.readthedocs.org/en/latest/intro.html#error-codes
 ignore=E501,E402
-exclude=.git,.pycharm_helpers
+exclude=.git,.pycharm_helpers,lib,src
```
cpennington: Why'd you add this exclusion?

course_discovery/locustfile.py
```diff
@@ -52,6 +57,9 @@ class CourseDiscoveryApiLocust(HttpLocust):
     def __init__(self):
         super(CourseDiscoveryApiLocust, self).__init__()
         self.client = LocustEdxRestApiClient(
-            config.COURSE_DISCOVERY_API_URL,
+            '%s%s' % (
```
cpennington: Nit: prefer `.format` to `%`.

cpennington:
High level question: Should we just be using https://docs.python.org/2/library/configparser.html?

cpennington:
Another high level question: Should whatever settings solution we end up with allow for specific values to be either injected via the commandline or environment variables, so that running the tests against a local server (for testing) doesn't require editing the yaml files?

cpennington:
(Or, put another way, what's the plan for loadtests against different environments? Have a separate set of yaml files somewhere that points to the loadtest environment?)

.pep8
pwnage101: In my local virtualenv, lib and src directories contained python files which failed the pep8 tests.  Various dependent libraries installed in the virtualenv, such as locust itself, contain poorly indented code, etc.

cpennington: Ah. I didn't think pep8 would pick up stuff in the virtualenvs. Was the virtualenv inside the repo?

pwnage101: The special virtualenv dirs are not checked-into the repo, but they do live in the repo if that's where the virtualenv gets created.

Having the load tests reorganized could allow us to point pep8 at them without special excludes (PERF-357 is this exactly).

pwnage101:
> High level question: Should we just be using
> https://docs.python.org/2/library/configparser.html?

AFAICT, configparser syntax supports strings only, but the settings files in this PR make use of numbers (integers), booleans, and lists of strings.  Which isn't to say that environment variables supported anything more than strings, but I consider these extra types an improvement since they help clarify and enforce data structure.

To be totally clear, I'm not dead set on yaml.  json was no-go because comments are unsupported, and honestly I hadn't considered much else.

pwnage101:
blehh, I totally skimmed over the RawConfigParser.getInt and getBoolean functions, so configparser does support integer and booleans.  No lists still, i think, but comma-delimited strings could become standardized via extension of the ConfigParser class.

cpennington:
To be clear, I'm not strictly opposed to using a yaml file either. I just tend to think the less we have to write ourselves for this sort of thing, the better.

pwnage101:
>Another high level question: Should whatever settings solution we end up with allow for specific values to be either injected via the commandline or environment variables, so that running the tests against a local server (for testing) doesn't require editing the yaml files?
>(Or, put another way, what's the plan for loadtests against different environments? Have a separate set of yaml files somewhere that points to the loadtest environment?)

One of my goals is to eliminate as many non-settings-file sources of input parameters as possible, so that the contents of the file can be used to determine how exactly the test was configured, just short of the code itself.

Given the code in the current PR, only one settings file path for a given test is acknowledged, so supporting multiple environments would require using symlinks for indirection (e.g. `settings_files/lms.py` -> `settings_files/<environment>/lms.py`), or just a straight copy (e.g. `cp settings_files/<environment>/lms.py settings_files`).

Regarding specifying custom settings files---I would like to do that, but there is no obvious way to pass custom options via locust command line arguments.  Also peeked into https://github.com/locustio/locust/blob/master/locust/main.py to see if there was an easy and reliable way to wrap the locust command to add custom command line arguments, but again nothing obvious to me.  I'm tempted to submit an issue to locustio/locust to start that discussion.

maxrothman:
Can you give me a little context on this? Does this relate to the configuration repo at all?

benpatterson:
@maxrothman I had advised @pwnage101 that you'd be a good reviewer for load test changes as someone that's made improvements on this repo and a member of the devops team fwiw

cpennington:
> One of my goals is to eliminate as many non-settings-file sources of input parameters as possible, so that the contents of the file can be used to determine how exactly the test was configured, just short of the code itself.

Is that the explicit goal? Or is the explicit goal repeatability, and this is the strategy to get to it? I guess I'm wondering whether we could allow more options for overriding, and just make sure that any reporting that happens during the test also includes the settings that were used (and, optimally, and exact commandline/config file to reproduce the test settings).

pwnage101:
Indeed, my explicit goal is repeatability.  Actually the code in this PR already spits out the inputs:

```
$ locust --host=https://pwnage101.sandbox.edx.org -f lms --no-web --clients=20 --hatch-rate=2
[2016-07-25 09:37:24,143] carbon/INFO/helpers.settings: using settings file: /home/tsankey/tmp/edx-load-tests/settings_files/lms.yml
[2016-07-25 09:37:24,145] carbon/INFO/helpers.settings: loaded the following settings:
{'COURSE_DATA': 'demo_course',
 'COURSE_ID': 'edX/DemoX/Demo_Course',
 'LOCUST_MAX_WAIT': 15000,
 'LOCUST_MIN_WAIT': 7500,
 'LOCUST_TASK_SET': 'LmsTest'}
[2016-07-25 09:37:24,145] carbon/INFO/locust.main: Starting Locust 0.7.3
[2016-07-25 09:37:24,145] carbon/INFO/locust.runners: Hatching and swarming 20 clients at the rate 2 clients/s...
[...]
```

Besides command line arguments (requires locust patch AFAICT) and shell environment variables, I'm not aware of any override mechanism which makes sense.  If it reads environment variables, I could just make it print a big warning that environment variables are being used.

pwnage101:
@maxrothman this is not related to the configuration repo, as it is not concerned with the edx platform directly. It's a change to the way individual load tests are configured.  See PERF-352 for more info.

pwnage101:
force-pushed the following updates:

* fix python syntax bugs
* use str.format instead of %
* make sure all locustfile.py files are not executable
* in helpers/settings.py use __name__ instead of __file__ for logger name

cpennington:
Ok, glad to see it printing out the config values.

Up to you how you want to proceed, at the moment. I'm fine with any of the solutions proposed so far, including the code as written.

course_import/locustfile.py
```diff
@@ -35,19 +27,23 @@ def on_start(self):
...
+        for i in xrange(settings.data['NUM_PARALLEL_COURSES']):
             self.create_course(i)
 
     def login(self):
         "Log in to CMS."
 
-        if BASIC_AUTH_CREDENTIALS:
-            self.client.auth = BASIC_AUTH_CREDENTIALS
+        if 'BASIC_AUTH_USER' in settings.data and \
+           settings.data['BASIC_AUTH_USER'] is not None:
```
cpennington: Nit: maybe just `if settings.data.get('BASIC_AUTH_USER') is not None`?

ecommerce/locustfile.py
```diff
@@ -1,10 +1,30 @@
...
+    'ECOMMERCE_ORDER_OFFSET',
+    'ECOMMERCE_ORDER_PREFIX',
+    'CYBERSOURCE_SECRET_KEY',
+    'LOCUST_TASK_SET',
+])
+
+# cleanup settings
+if settings.data.get('PAID_SKU') is None:
+    raise RuntimeError('A SKU corresponding to a paid course seat is' +
+                       'required to run load tests.')
```
cpennington: You don't need, this, right, since you're marking `PAID_SKU` as a required setting?

helpers/auto_auth_tasks.py
```diff
@@ -21,8 +14,13 @@ def __init__(self, *args, **kwargs):
...
         super(AutoAuthTasks, self).__init__(*args, **kwargs)
-        if BASIC_AUTH_CREDENTIALS:
-            self.client.auth = BASIC_AUTH_CREDENTIALS
+
+        basic_auth_credentials = (
+            settings.data.get('BASIC_AUTH_USER', None),
+            settings.data.get('BASIC_AUTH_PASS', None),
+        )
+        if basic_auth_credentials[0] is not None:
+            self.client.auth = basic_auth_credentials
```
cpennington: In other parts of the code, you've used the pattern:

    if settings.data.get('BASIC_AUTH_USER') is not None:
        self.client.auth = ...

It would be nice to be consistent across the tests.

helpers/settings.py
```diff
@@ -0,0 +1,55 @@
...
+
+    Anywhere you need to use the settings data, make sure the settings module
+    is imported, then use:
+
+      settings.data['SOMETHING']
+
+    """
+    global data
+
+    if 'data' in vars():
```
cpennington: This makes me a bit uncomfortable. You could get the same results by just putting `data` in the global scope, set to `None`, right?

pwnage101: Sorry, could you elaborate a bit more on this note?  Are you arguing to remove this check and just define the data variable immediately?

cpennington: I generally lean away from using `vars` to pull out the variables in scope. I'd prefer to let python do it for me. I'm thinking something like:

    data = None
    
    def init(...):
        global data
        
        if data is not None:
            raise RuntimeError(...)
        data = {...}

pwnage101: This does present a corner case: when the settings file is empty *and* settings gets initialized twice.  Fortunately this case is benign, so I'll incorporate it since it improves readability.

testing:
```
>>> from helpers import settings
>>> settings.data is None
True
>>> settings.init('lms.locustfile')
>>> settings.data
{'COURSE_ID': 'edX/DemoX/Demo_Course', 'COURSE_DATA': 'demo_course', 'LOCUST_MAX_WAIT': 15000, 'LOCUST_MIN_WAIT': 7500, 'LOCUST_TASK_SET': 'LmsTest'}
```

pwnage101:
Latest comments have been addressed.

Today I was finally able to test the discussion_api tests, and that was successful.  I wonder if I should keep testing the rest?  It isn't always straightforward to prepare the tests, but I am learning quite a bit by necessity.

pwnage101:
I've decided to skip testing the Programs and Credentials load tests because they require a change (request JWTs rather than generate them) which is out of the scope of this PR.  What's left to test:

*  course_discovery
*  course_import
*  csm
*  ecommerce

pwnage101:
Alright, seeking reviews again.  In summary:

* improved some of the default settings
* reverted settings implementation in the following load tests:
 * credentials - could not sufficiently test this service due to deprecated auth method (will address in PERF-361), so I'm not going to touch this load test.
 * ecommerce - same as above (see PERF-361).
 * programs - this service is being retired, not going to do anything with this load tests (see ECOM-4418).

As it is, the csm load test crashes because when it tries to load the `lms.envs` module from edx-platform, it first tries to load the local lms module (the LMS load tests) and settings gets doubly loaded which raises an exception.  I'd still like to merge my changes, then this will soon be addressed by reorganizing the load tests into a subdirectory via another PR.

@cpennington @tobz @maxrothman

cpennington:
:+1:

tobz:
👍 once merged w/ master tests pass.

PR #60, opened 2016-08-01
TNL-5023: Add seeding and testing of large thread.

robrap:
### Description
 
[TNL-5023](https://openedx.atlassian.net/browse/TNL-5023)

TNL-5023: Forums Performance: Investigate Large Threads

Here is a first pass at seeding/testing large threads.  

This code adds a Locust Task Set for:

1. Seeding a large thread.
2. Reading the large thread.

The seeding was implemented as a Locust Task Set which can be run as a separate test.  This enables the seeding to take advantage of the parallelism of Locust, and we can also compare seeding times.

Here's how to seed:
```
BASIC_AUTH_USER=XXX BASIC_AUTH_PASSWORD=YYY COURSE_ID=course-v1:edX+DemoX+Demo_Course LOCUST_TASK_SET=SeedForumsTasks LOCUST_MIN_WAIT=1000 LOCUST_MAX_WAIT=1000 locust -c 10 -r 1 -n 4500 -H https://discussions.sandbox.edx.org --no-web
````

Testing against a large thread just does repeated reads:
```
BASIC_AUTH_USER=XXX BASIC_AUTH_PASSWORD=YYY LARGE_TOPIC_ID=i4x-edx-eiorguegnru-course-foobarbaz LARGE_THREAD_ID=579f8b6948ee34412e000078 COURSE_ID=course-v1:edX+DemoX+Demo_Course LOCUST_TASK_SET=ForumsTasks LOCUST_MIN_WAIT=1000 LOCUST_MAX_WAIT=1000 locust -c 10 -r 1 -H https://discussions.sandbox.edx.org --no-web
```

I'll drops some comments in the code as well.

### Reviewers (any 2 of the following):
- [x] @tobz 
- [x] @pwnage101 
- [ ] @ormsbee 
- [ ] @dianakhuang 

FYI: @andy-armstrong @bjacobel 

### Post-review
- [x] Squash commits

### DIFF
--- a/lms/forums.py
+++ b/lms/forums.py
@@ -1,16 +1,15 @@
 from collections import deque
+import logging
+import os
 import random
 import string
 
+from lazy import lazy
+from lms import LmsTasks
 from locust import task
 
-from lms import LmsTasks
-
-
-# keep track of thread ids created during a test, so they can be used to formulate read requests and replies.
-# a deque is used instead of a list in order to enforce maximum length.
-_thread_ids = deque(maxlen=1000)
-
+logging.basicConfig(level=logging.INFO)
+LOGGER = logging.getLogger(__name__)
 
 _dummy_chars = string.lowercase + '      '
 
@@ -23,79 +22,72 @@ def _dummy_text(minlen, maxlen):
     return ''.join(random.choice(_dummy_chars) for _ in xrange(minlen, random.randrange(minlen + 1, maxlen)))
 
 
-class ForumsTasks(LmsTasks):
+class BaseForumsTasks(LmsTasks):
     """
-    Simulate traffic for endpoints in lms.djangoapps.django_comment_client.base and .views
+    Base class for Forums (LMS) TaskSets.
 
-    Traffic Distribution on courses.edx.org (last 7d as of 2014-02-24):
+    This class supports environment-based configuration to override default
+    values for the following:
 
-    /django_comment_client.base.views:create_comment            17205     1.73%
-    /django_comment_client.base.views:create_sub_comment        8489      0.85%
-    /django_comment_client.base.views:create_thread             31984     3.22%
-    /django_comment_client.base.views:delete_comment            434       0.04%
-    /django_comment_client.base.views:delete_thread             368       0.04%
-    /django_comment_client.base.views:endorse_comment           1032      0.10%
-    /django_comment_client.base.views:flag_abuse_for_comment    69        0.01%
-    /django_comment_client.base.views:flag_abuse_for_thread     62        0.01%
-    /django_comment_client.base.views:follow_thread             2895      0.29%
-    /django_comment_client.base.views:openclose_thread          27        0.00%
-    /django_comment_client.base.views:pin_thread                177       0.02%
-    /django_comment_client.base.views:un_flag_abuse_for_comment 35        0.00%
-    /django_comment_client.base.views:un_flag_abuse_for_thread  36        0.00%
-    /django_comment_client.base.views:un_pin_thread             70        0.01%
-    /django_comment_client.base.views:undo_vote_for_comment     285       0.03%
-    /django_comment_client.base.views:undo_vote_for_thread      490       0.05%
-    /django_comment_client.base.views:unfollow_thread           567       0.06%
-    /django_comment_client.base.views:update_comment            2483      0.25%
-    /django_comment_client.base.views:update_thread             1891      0.19%
-    /django_comment_client.base.views:upload                    1051      0.11%
-    /django_comment_client.base.views:users                     13678     1.38%
-    /django_comment_client.base.views:vote_for_comment          4691      0.47%
-    /django_comment_client.base.views:vote_for_thread           7012      0.71%
-    /django_comment_client.forum.views:followed_threads         7910      0.80%
-    /django_comment_client.forum.views:forum_form_discussion    184475    18.55%
-    /django_comment_client.forum.views:inline_discussion        150204    15.11%
-    /django_comment_client.forum.views:single_thread            547876    55.10%
-    /django_comment_client.forum.views:user_profile             8765      0.88%
+    * LARGE_TOPIC_ID: Topic id for the large thread.
+    * LARGE_THREAD_ID: Thread id for the large thread.
+    See subclasses for usage.
+
     """
 
-    @task(19)
-    def forum_form_discussion(self):
-        """
-        Visit the discussion tab.
-        """
-        self.get('discussion/forum', name="forums:forum_form_discussion")
+    # class variables used to share across locust users.
+    # track the large thread tuple (topic_id, thread_id).
+    _large_thread = None
+    _all_topic_ids = None
 
-    @task(5)
-    def search_topic(self):
+    def random_topic_id(self):
         """
-        Load a randomly-selected topic from the discussion tab sidebar.
-        Uses ES, but not actually a search from the user perspective.
+        randomly choose the topic_id (aka discussion_id or commentable_id)
         """
-        self.get(
-            'discussion/forum/search',
-            params={
-                'commentable_ids': self.course_data.discussion_id,
-                'page': '1',
-            },
-            name='forums:search_topic',
-        )
+        return random.choice(self.topic_ids())
 
-    @task(15)
-    def inline_discussion(self):
-        """
-        Load a randomly-selected inline discussion.
-        """
-        self.get(
-            'discussion/forum/{}/inline'.format(self.course_data.discussion_id),
-            params={'page': '1'},
-            name='forums:inline_discussion',
-        )
+    def topic_ids(self):
+        # use Discussion API to load topics
+        if not BaseForumsTasks._all_topic_ids:
+            url = '/api/discussion/v1/course_topics/{}/'.format(self.course_key)
+            response = self.client.get(url, verify=False, name='forums:get_topics')
+            response_json = response.json()
 
-    @task(4)
-    def create_thread(self):
+            BaseForumsTasks._all_topic_ids = []
+            for topic in response_json['non_courseware_topics']:
+                BaseForumsTasks._all_topic_ids.append(topic['id'])
+            for topic in response_json['courseware_topics']:
+                for child in topic['children']:
+                    BaseForumsTasks._all_topic_ids.append(child['id'])
+
+        return BaseForumsTasks._all_topic_ids
+
+    @lazy
+    def large_topic_id(self):
         """
-        create a new thread in a randomly-selected topic
+        The topic id of the large thread.
+        """
+        if BaseForumsTasks._large_thread:
+            return BaseForumsTasks._large_thread[0]
+        else:
+            return os.getenv('LARGE_TOPIC_ID', None)
+
+    @lazy
+    def large_thread_id(self):
+        """
+        The thread id of the large thread.
+        """
+        if BaseForumsTasks._large_thread:
+            return BaseForumsTasks._large_thread[1]
+        else:
+            return os.getenv('LARGE_THREAD_ID', None)
+
+    def create_thread(self, topic_id, name):
+        """
+        Create a new thread in the topic_id.
+
+        Returns:
+            Tuple: (topic_id, thread_id)
         """
         thread_data = {
             'body': _dummy_text(100, 2000),  # NB size range not based on actual data.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: lms/forums.py [hunk 1]
```diff
@@ -1,16 +1,24 @@
...
 
+logging.basicConfig(level=logging.INFO)
+LOGGER = logging.getLogger(__name__)
 
 # keep track of thread ids created during a test, so they can be used to formulate read requests and replies.
 # a deque is used instead of a list in order to enforce maximum length.
 _thread_ids = deque(maxlen=1000)
 
+# track the large thread tuple (topic_id, thread_id)
+_large_thread = []
```
robrap: Reviewers: Is it fine to just continue this pattern used for _thread_ids?  Should these be moved to be class variables?

lms/forums.py
```diff
@@ -172,3 +180,139 @@ def followed_threads(self):
...
+                'resp_skip': skip_limit[0],
+                'resp_limit': skip_limit[1],
+            },
+            name='forums:large_thread',
+        )
+
+
+class SeedForumsTasks(LmsTasks):
+
+    def create_thread(self, topic_id):
```
robrap: Reviewers: Some of this code is partly redundant with ForumsTasks.  Python question: if I want to share, since these are using `self.post()`, would I just add in a ForumsBaseTasks to inherit from?

lms/forums.py
```diff
@@ -172,3 +180,139 @@ def followed_threads(self):
...
+            'body': _dummy_text(100, 2000),  # NB size range not based on actual data.
+        }
+        response = self.post(
+            'discussion/threads/{}/reply'.format(thread_id),
+            data=thread_data,
+            name='forums:create_response',
+        )
+
+        # clump comments on fewer or responses
+        if not _large_thread_response_ids or random.randint(1, 10) > 9:
```
robrap: @tobz: How about this?
```
if not _large_thread_response_ids or random.randint(1, len(_large_thread_response_ids)) <= 1
    ...
```

tobz: What happens if there are no threads? Does `not _empty_list_here` evaluate to a truthy expression?

lms/forums.py
pwnage101: Fine by me.  But if you do move it into the class, remember to declare it global.

Also, why is `_large_thread` of type list if you never append more than one element or reference anything except the zeroth index?  AFAICT, you can initialize it as `None` and refer to it as a tuple rather than a list.

lms/forums.py
robrap: @tobz: Yes.  If it is empty, then the first response will get added and we don't make it to the second half of the or.

lms/forums.py
robrap: @pwnage101: I feel like that caused an issue, but not sure when, so let me fix.

lms/forums.py
```diff
@@ -172,3 +180,139 @@ def followed_threads(self):
...
+            'body': _dummy_text(100, 2000),  # NB size range not based on actual data.
+        }
+        response = self.post(
+            'discussion/threads/{}/reply'.format(thread_id),
+            data=thread_data,
+            name='forums:create_response',
+        )
+
+        # clump comments on a few responses
+        if not _large_thread_response_ids or random.randint(1, len(_large_thread_response_ids)) <= 1:
```
tobz: Can we just go ahead and change `not _large_thread_response_ids` to something like `len(_large_thread_response_ids) < RESPONSES_ABOVE_THE_FOLD` or something?  I don't really care what the constant name is, but I was mostly thinking in the sense of... how many responses are visible for most users without having to scroll down at all?

I think we're golden here if we have both  the "always add the first N responses" and "gradually comment less and less on newer responses" behavior.

lms/forums.py
```diff
@@ -172,3 +180,139 @@ def followed_threads(self):
...
+                for child in topic['children']:
+                    all_topic_ids += '"{}",\n'.format(child['id'])
+
+            new_thread = self.create_thread(large_thread_topic_id)
+            _large_thread.append(new_thread)
+
+            LOGGER.info('To run ForumsLargeThreadTasks, use:\nLARGE_TOPIC_ID={} and LARGE_THREAD_ID={}'.format(
+                    new_thread[0], new_thread[1])
+            )
+            LOGGER.info('To run ForumsTasks, update course_data discussion_ids to:\n{}'.format(all_topic_ids))
```
pwnage101: How long does seeding take for a large thread?  Is this seeding process now required for ForumsTasks (not for just ForumsLargeThreadTasks), as implied by the output message?

lms/forums.py
pwnage101: I'm okay with leaving this as is, since the same-named tasks aren't exactly the same code.   If I were to ask for anything more, I'd probably ask that you somehow improve LmsTasks to reduce boilerplate code across all the tasks.

lms/forums.py
robrap: I needed to add a `global` to make it work that I must have missed my first attempt.

lms/forums.py
robrap: @pwnage101: I did end up refactoring this before seeing your comment.  I moved shared code to a Base class.

I'm not sure what boilerplate code you are looking for in LmsTasks.  Code that affects this Forums code?  Otherwise, I'd rather just concentrate on the Forums code for this PR.

lms/forums.py
robrap: @tobz: I "think" I sort of understand what you are asking for, but not certain.  Do you want a bunch of comments guaranteed on the oldest response?  If so, I think this would require tracking the number of comments added, and then adding to the response ids after we have that number of comments first.  It feels a bit complicated, but may be needed.  Can we start with what I have and then see where it gets us?

lms/forums.py
robrap: I know.  This wasn't clear because I was dumping course_data discussion ids, because this wasn't done anywhere else and I wanted to give someone the ability to fix the test (which is otherwise broken for LMS in general).  You don't actually require the large thread for lms tests, and it wasn't originally integrated.

I did add some integration where if you provide the large topic/thread, then LMS tests will also do some of its reads against that large thread.  Right now the ration is 45:10 (any thread:large thread).  We could adjust.  I just picked anything to get integration started.

robrap:
@tobz @pwnage101: I made a pass at fixes.  Let me know if you have any required changes or how to get thumbs.  Thanks!

lms/forums.py
```diff
@@ -172,3 +267,83 @@ def followed_threads(self):
...
+
+
+class ForumsLargeThreadTasks(BaseForumsTasks):
+
+    @task(1)
+    def large_thread(self):
+        """
+        Read the large thread.
+        """
+        super(ForumsTasks, self).large_thread()
```
pwnage101: s/ForumsTasks/ForumsLargeThreadTasks/

behavior is the same, but it's more understandable.

robrap: @pwnage101: Just let me know when you are done with your pass and I'll push changes.  Thanks.

lms/forums.py
```diff
@@ -23,7 +31,103 @@ def _dummy_text(minlen, maxlen):
...
+
+        new_thread = (topic_id, response.json()['id'])
+        return new_thread
+
+    def large_thread(self):
+        """
+        Read the large thread.
+        """
+        discussion_id = self.large_topic_id
+        thread_id = self.large_thread_id
```
pwnage101: lines 85,86---no reason to do this really

lms/forums.py
```diff
@@ -133,8 +223,20 @@ def single_thread(self):
...
+    def large_thread(self):
+        """
+        Read the large thread.
+        """
+        if self.large_topic_id and self.large_thread_id:
+            super(ForumsTasks, self).large_thread()
+
+        else:
+            # Fall back to regular read if not configured for long thread test
+            self.single_thread
```
pwnage101: does the fallback mode actually call self.single_thread, or does it just return a function object?

robrap: Interesting.  This should be a function call.  However, I thought I tested this and saw single_threads appear in locust output, so I'm not sure what happened.

Either way, I will update and retest.  Thanks for catching this.

lms/forums.py
pwnage101: high level question: why even have a ForumsLargeThreadTasks task set at all if the ForumsTasks can cover large threads now?

pwnage101:
high level question: why even have a ForumsLargeThreadTasks task set at all if the ForumsTasks can cover large threads now?

pwnage101:
@robrap 
> I'm not sure how to make this configurable.

In the usual way: shell variables.

Sorry I'm being critical, but why would you want the ability to crank up the ratio of one task (large threads)?  Do the results of that tell us anything new?  Why not just run the ForumsTasks with the ratios in this PR all the time?  Running the seed tasks would "configure" whether to test large threads or not.

Just making sure we don't add code where it isn't needed, but please correct me if I'm wrong.

robrap:
@pwnage101: It's a reasonable question and what you say makes sense.  I think this is just an artifact of how I built this, where integration to the other tests at all was a last minute idea.

I'm going to adjust the ratio to 30 single : 25 large and remove the Large Thread Task Set.  I have a future task to look at load numbers during our peak, and may adjust again at that point.

robrap:
@pwnage101: On a side note, Locust complains when you call interrupt() on the top-level Task Set, so I'm not sure of the best way to handle that for Task Sets that are sometimes run directly.  I guess instead we'd need LMS Task Sets with single sub-Task Sets, so it remains a sub-Task Set.

robrap:
@pwnage101 @tobz: I added final fixes and adjusted weightings based on my research.  This is ready for review.  I haven't actually run the latest latest changes, but I plan to shortly with @tobz for our current round of load testing against the sandbox.

robrap:
@pwnage101: Do you think we should move away from the hard coded course_data, and use the same call during setup of ForumsTasks to determine discussion ids?  I may do this is a separate PR.  Thoughts?

lms/forums.py
```diff
@@ -1,16 +1,24 @@
...
+LOGGER = logging.getLogger(__name__)
 
 # keep track of thread ids created during a test, so they can be used to formulate read requests and replies.
 # a deque is used instead of a list in order to enforce maximum length.
 _thread_ids = deque(maxlen=1000)
 
+# track the large thread tuple (topic_id, thread_id)
+_large_thread = None
+# track responses on the large thread
+_large_thread_response_ids = deque(maxlen=100)
```
pwnage101: I think there's no reason to declare `_large_thread` to be global in the SeedForumsTasks class below because it's already global to this module.   I think what I was talking about earlier was to move declaration and initialization of both of these variables into the various classes below as global member variables (e.g. `self._large_thread`).

I suppose you'd declare `self._large_thread` in BaseForumsTasks and `self._large_thread_response_ids` in SeedForumsTasks.

pwnage101: Okay, so I did a bit more research and it turns out class member variables can't be declared global anyway.  They're already sort of global when referenced like a module, e.g. `BaseForumsTask._large_thread`.  I'll leave it up to you how to move forward, but at least remove the global declaration and make sure that is a noop.

robrap:
@pwnage101 @tobz: I did a big refactor in the last commit.
1. Moved to class variables.
2. Removed discussion_ids (topic_ids) from course_data and dynamically load at the start of a test.  This make it much simpler to run the ForumsTasks against different courses in different environments.

@pwnage101: FYI: Moving to class variables, I no longer need the `global` call.  However, it was originally needed to avoid the following error:
```
UnboundLocalError: local variable '_large_thread' referenced before assignment
```

robrap:
@tobz @pwnage101: Sorry for the late commit (see above).  I'd like to see if I can get this merged before end of iteration (tomorrow morning), but that may not be possible.

FYI: @dianakhuang

lms/forums.py
```diff
@@ -127,40 +155,31 @@ def create_comment(self, response_id):
...
+    * LARGE_TOPIC_ID: Topic id for the large thread.  If blank, all reads
+        will be against normal sized threads.
+    * LARGE_THREAD_ID: Thread id for the large thread.  If blank, all reads
+        will be against normal sized threads.
+
+    """
+
+    # keep track of thread ids created during a test, so they can be used to formulate read requests and replies.
+    # a deque is used instead of a list in order to enforce maximum length.
+    _thread_ids = deque(maxlen=1000)
```
pwnage101: can you add another comment about the intention of using this as a global variable, rather than an instance var?

robrap: Done.

lms/forums.py
```diff
@@ -23,50 +22,174 @@ def _dummy_text(minlen, maxlen):
...
+            response_json = response.json()
+
+            BaseForumsTasks._all_topic_ids = []
+            for topic in response_json['non_courseware_topics']:
+                BaseForumsTasks._all_topic_ids.append(topic['id'])
+            for topic in response_json['courseware_topics']:
+                for child in topic['children']:
+                    BaseForumsTasks._all_topic_ids.append(child['id'])
+
+        return BaseForumsTasks._all_topic_ids
```
tobz: This is great.  I think this will solve the problem I was having where the hardcoded `course` topic ID was making things funky.  It's also a nice sanity check that we make the discussion service tell us what it thinks it has.

lms/forums.py
```diff
@@ -97,32 +220,17 @@ def create_thread(self):
...
-        _thread_ids.append((discussion_id, response.json()['id']))
-
-    @task(55)
+    @task(36)
     def single_thread(self):
         """
         Read a specific thread.
         """
-        if not _thread_ids:
+        if not ForumsTasks._thread_ids:
```
tobz: Why can't this be `self._thread_ids`? (I'm not asking in the sense of "why didn't you do that?" but "is that not a valid way?")

robrap: I tried and it didn't work. I'm new to Python, so not sure whether this should be a valid way to get to a class variable and I'm just doing something wrong?

pwnage101: Quoting Robert Raposa (2016-08-10 18:34:07)
> I tried and it didn't work. I'm new to Python, so not sure whether this
> should be a valid way to get to a class variable and I'm just doing something
> wrong?

IIRC, you're using `ForumsTasks._thread_ids` because it is shared between
instances of the class, and that's the one that gets defined in on_start.  If
you use the instance variable `self._thread_ids`, then it should throw
"undefined".

Troy

tobz:
👍

pwnage101:
:+1:

PR #62, opened 2016-08-09
MA-2696: load test new index for unread_comment_count

wajeeha-khalid:
[MA-2696](https://openedx.atlassian.net/browse/MA-2696)

Run load tests on a new index to calculate `unread_comment_count`.
`db.contents.ensureIndex({ comment_thread_id: 1, author_id: 1, created_at: 1 }, { background: true })`

### DIFF
--- a/discussions_api/db/create_indexes.js
+++ b/discussions_api/db/create_indexes.js
@@ -259,6 +259,17 @@ var contentsIndexes = [
         "name" : "author_id_1_course_id_1",
         "ns" : "comments-prod.contents",
         "background" : true
+    },
+    {
+        "v" : 1,
+        "key" : {
+            "comment_thread_id" : 1,
+            "author_id" : 1,
+            "created_at" : 1
+        },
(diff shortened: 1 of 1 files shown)


pwnage101:
@wajeeha-khalid I thought there was one index added and one removed (recalling a previous conversation we had), but this change only adds an index.  Maybe I was thinking about a different change?

wajeeha-khalid:
@pwnage101 Thats true; I have not added remove index to this PR yet (or more specifically `discussions_api/db/create_indexes.js`); because we intend to keep old index as is (for some time) in case we want to rollback.
So, I explicitly asked devops to add new index and remove old one on server deployment ticket. Which means the tests were run according to what you assume; one index added one removed; and the load test server has the same state.

Reference: https://openedx.atlassian.net/servicedesk/customer/portal/3/DEVOPS-4656

pwnage101:
:+1:

feanil:
Closing due to age.

PR #63, opened 2016-08-11
Load tests for catalog search

rlucioni:
Tests for ECOM-2668. @clintonb please review.

FYI @edx/ecommerce


course_discovery/locustfile.py
```diff
@@ -1,65 +1,116 @@
...
     def get_catalog_courses(self):
-        """ Retrieve all courses associated with a catalog. """
-        self.client.catalogs(self.CATALOG_ID).courses.get()
+        """Retrieve all courses associated with a catalog."""
+        self.client.catalogs(self.catalog_id).courses.get()
+
+    @task(10)
+    def list_courses_with_query(self):
+        """Query the courses."""
+        query = {
```
clintonb: This can be simplified: `query = 'organizations:(MITx OR HarvardX)'`.

clintonb:
👍

course_discovery/locustfile.py
rlucioni: Nice. Updated.

PR #64, opened 2016-08-11
improve docs related to settings

pwnage101:
Grok'd the READMEs for anything suggesting that we use the old style settings (environment variables) and updated it. Also:

* freenode -> slack
* show how to install custom requrements
* use something other than `programs` as an example (programs is being retired).


pwnage101:
@robrap (since we just talked about this)

pwnage101:
also @tobz

just looking for a single thumb

tobz:
👍

PR #65, opened 2016-08-12
settings files ignored by git + fix rebase bug

pwnage101:
following up on PERF-352 / #58 :(

Make sure the settings files are ignored by git because we don't want people committing secrets.

Also, fix a rebase bug in the lms tests.  I forgot to include two more settings.

@rlucioni @robrap


rlucioni:
👍

robrap:
👍

PR #66, opened 2016-08-23
WikiViewTask

efischer19:
@pwnage101 Thanks for all the work that's been done here recently - it's so easy to run simple load tests now!

I'm mainly putting this up to maintain a record of what I did, so the next time can happen even more easily. If you think anyone would benefit from the `WikiViewTask` I created, we can merge that, otherwise I'll close the PR without merging.

### DIFF
--- a/settings_files/lms.yml.sandbox-example
+++ b/settings_files/lms.yml.sandbox-example
@@ -1,0 +1,23 @@
+---
+# Course ID that we're running these tests against.  The default
+# targets a course which currently exists on {your_sandbox}.sandbox.edx.org.
+COURSE_ID: course-v1:edX+DemoX+Demo_Course
+
+# The name of the course data variable which corresponds to the
+# COURSE_ID specified above.  Must be defined in the lms.course_data
+# module.
+COURSE_DATA: demo_course
+
+# Run the specified TaskSet (must be imported into the lms/locustfile.py
+# file):
+LOCUST_TASK_SET: LmsTasks
+
+# Minimum/Maximum waiting time between the execution of locust tasks:
+LOCUST_MIN_WAIT: 500
+LOCUST_MAX_WAIT: 1500
+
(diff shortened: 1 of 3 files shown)


settings_files/lms.yml.example
```diff
@@ -1,7 +1,7 @@
 ---
 # Course ID that we're running these tests against.  The default
 # targets a course which currently exists on courses-loadtest.edx.org.
-COURSE_ID: edX/DemoX/Demo_Course
+COURSE_ID: course-v1:edX+DemoX+Demo_Course
```
pwnage101: I fell into this one too: #57 

In an effort to keep things consistent, it's probably best to keep the defaults compatible with one target (the loadtest environment.  Perhaps you can include this change in a new file called `lms.yml.sandbox-example`?  In the same vein, i may rename the current .example files to .loadtest-example.  Thoughts?

pwnage101:
Thanks for your records!  I actually would like the settings file changes included.  Also if you think it's generally useful, include the wiki task with the low/high slug numbers pulled from settings.  I don't care if it isn't runnable by default, the user can bypass straight to it using the following setting:

```yaml
LOCUST_TASK_SET: WikiViewTask
```

efischer19:
@pwnage101 Settings file addressed, does this look okay to merge?

settings_files/lms.yml.sandbox-example
```diff
@@ -0,0 +1,23 @@
...
+COURSE_ID: course-v1:edX+DemoX+Demo_Course
+
+# The name of the course data variable which corresponds to the
+# COURSE_ID specified above.  Must be defined in the lms.course_data
+# module.
+COURSE_DATA: demo_course
+
+# Run the specified TaskSet (must be imported into the lms/locustfile.py
+# file):
+LOCUST_TASK_SET: WikiViewTask
```
pwnage101: Can you keep this LmsTasks? Otherwise, :+1:

PR #68, opened 2016-08-24
makefile works for all tests

pwnage101:
* also add make targets for the settings files
* also add "requirements" make taget

---

some examples:
```
$ make requirements
pip install -r requirements.txt --exists-action w
Downloading/unpacking edx-opaque-keys==0.3.2 (from -r requirements/base.txt (line 1))
  Downloading edx-opaque-keys-0.3.2.tar.gz
...
```

```
$ HOST=http://localhost:8009 make enrollment
locust --host=http://localhost:8009 -f enrollment
[2016-08-24 17:10:15,826] carbon/INFO/helpers.settings: using settings file: /home/tsankey/tmp/edx-load-tests/edx-load-tests/settings_files/enrollment.yml
[2016-08-24 17:10:15,827] carbon/INFO/helpers.settings: loaded the following settings:
...
```
```
$ make
see README.rst for usage details
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,34 @@
-course_discovery:
-	locust --host=$(HOST) -f course_discovery
+# Update this list when you add a new test.
+TESTS = \
+    course_discovery \
+    course_import \
+    credentials \
+    csm \
+    discussions_api \
+    ecommerce \
+    enrollment \
+    lms \
+    teams_discussion
 
-.PHONY: course_discovery
+help:
+	@echo 'see README.rst for usage details'
+
+requirements:
+	pip install -r requirements.txt --exists-action w
+
+# I don't understand why the .SECONDEXPANSION feature needs to be enabled to
+# use $@ (target reference) in prerequisites, but it does.  Please help to
+# figure out a cleaner solution.
+.SECONDEXPANSION:
+
+# If the settings file is missing, just copy the example settings.  If the
+# settings file already exists, this recipe will not be triggered. Your custom
+# settings should never not be overwritten.
+settings_files/%.yml: $$@.example
+	cp $< $@
+
+$(TESTS): settings_files/$$@.yml
+	locust --host=$(HOST) -f $@
+
+.PHONY: help requirements $(TESTS)
+.DEFAULT: help
--- a/README.rst
+++ b/README.rst
@@ -14,7 +14,7 @@ Next, install load testing requirements.
 
 .. code-block::
 
-    $ pip install -r requirements.txt
+    $ make requirements
 
 If the load test in question has additional requirements, install those too:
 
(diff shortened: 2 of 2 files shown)


pwnage101:
@rlucioni since you added the makefile, I assume you'd be interested.

Makefile
```diff
@@ -1,4 +1,34 @@
...
-.PHONY: course_discovery
+help:
+	@echo 'see README.rst for usage details'
+
+requirements:
+	pip install -r requirements.txt --exists-action w
+
+# I don't understand why the .SECONDEXPANSION feature needs to be enabled to
+# use $@ (target reference) in prerequisites, but it does.  Please help to
+# figure out a cleaner solution.
```
rlucioni: http://www.pyinvoke.org/ is one option for a cleaner solution. Invoke tasks are easier to understand than some of the magic going on in this Makefile.

Makefile
```diff
@@ -1,4 +1,34 @@
...
+.SECONDEXPANSION:
+
+# If the settings file is missing, just copy the example settings.  If the
+# settings file already exists, this recipe will not be triggered. Your custom
+# settings should never not be overwritten.
```
rlucioni: "never not"?

Makefile
pwnage101: Unless you have objections, I'm actually going to evaluate paver.  What I want to do is not that complicated, so I'm asserting that any paver limitations should be outweighed by the fact that we already use paver elsewhere.

rlucioni: @pwnage101 either way, I think using Python to run these tasks is the right way to go.

I don't think Paver was the right choice for devstack. I would encourage you to consider [Invoke](https://github.com/pyinvoke/invoke). Invoke is the successor to [Fabric](https://github.com/fabric/fabric), both of which have much broader communities than [Paver](https://github.com/paver/paver). [Paver's docs](http://paver.readthedocs.io/en/latest/) are also very broken. [Invoke's docs](http://www.pyinvoke.org/) are much better.

benpatterson: Not disputing @rlucioni 's points, but I would rather implement with `paver`. We already have `Make` and `paver` in our task-management space (and, really, there's `gulp` for JS). I think a deeper discussion on whether or not we want to adopt pyinvoke (and put the resources into porting over to it) might be worthwhile. It's been >1 year since we last visited the paver/pyinvoke/make discussion. but we are using paver on 2 or 3 repos right now and I'd rather not stray from that model without a more thorough discussion about our overall strategy, with plans, etc.

rlucioni: @benpatterson fair. FWIW, that discussion might benefit from having an Invoke use case to point to. This repo seems like a good place to experiment.

benpatterson: Hmm. I'd be open to it being on a branch, but I'd rather not have it merged up, if that is a fair enough middle-ground.

pwnage101: fwiw, here is a WIP branch https://github.com/edx/edx-load-tests/tree/pwnage101/reorganize-and-invoke

rlucioni: @pwnage101 nice. I'm fine with either leaving that on a branch or merging it. I think deciding where to go from here is up to you and @benpatterson.

pwnage101:
closing in favor of paver implementation in #71

PR #71, opened 2016-08-31
reorganize tests and implement paver

pwnage101:
I realize this is two different things; if necessary I can split into two different PRs.

**Reorganize tests**: all the tests are now located beneath the ``loadtests`` directory, and the canonical locust invocation changed to:

    locust --host=<hostname> -f loadtests/<test-name>

So this would be a breaking change since ``-f lms`` wouldn't work anymore.

**Implement paver**: refer to the updated README.rst for detailed usage, but here's some output:

    % paver settings --loadtest-name=lms
    ---> pavement.settings
    cp settings_files/lms.yml.example settings_files/lms.yml
    NOTICE: settings for the lms load test can be changed by modifying
    the following file:
    
      settings_files/lms.yml

The primary motivation for doing this is so we can better insulate the locust ansible role from repository semantics.


pwnage101:
@rlucioni @tobz 

FYI @benpatterson

loadtests/course_discovery/locustfile.py
```diff
@@ -1,3 +1,9 @@
+import os
+import sys
+
+# due to locust sys.path manipulation, we need to re-add the project root.
+sys.path.append(os.path.dirname(os.path.dirname(os.path.dirname(__file__))))
```
rlucioni: Can you show me the relevant sys.path manipulation? Can we make a change to the project structure to avoid the need for this in every single locustfile?

pavement.py
```diff
@@ -0,0 +1,81 @@
+from __future__ import print_function
```
rlucioni: Out of scope for this, but I'd be interested in seeing what it would take to use Python 3 in this repo.

pavement.py
```diff
@@ -0,0 +1,81 @@
...
+import os
+import os.path
+from paver.easy import task, sh, cmdopts
+from optparse import make_option
```
rlucioni: optparse is deprecated as of 2.7. argparse should be used instead. Is this necessary to use paver?

pavement.py
```diff
@@ -0,0 +1,81 @@
...
+
+# tests are located under this directory
+SETTINGS_DIR = 'settings_files'
+
+# list of valid environments:
+VALID_ENVIRONMENTS = [
+    'loadtest',  # aka courses-loadtest.edx.org
```
rlucioni: This is pretty specific to us/edx.org. Do we know if anyone outside the building uses these tests?

pavement.py
```diff
@@ -0,0 +1,81 @@
...
+    make_option(
+        '-f', '--force', action='store_true', default=False,
+        help='force the overwrite of pre-existing settings'),
+])
+def settings(options):
+    """ensure a settings file is present"""
+    # NOTE: Some tests may not implement helpers.settings.  Those tests are not
+    # supported by this task.
+
+    # There no way in paver to make an option required.  There's also no way to
```
rlucioni: 😢

pavement.py
```diff
@@ -0,0 +1,81 @@
...
+    # mix positional arguments with optional arguments without introducing our
+    # own argument parser.  Here we manually gate for required options:
+    if options.get('loadtest_name') is None:
+        raise RuntimeError('missing required option: --loadtest-name=')
+
+    desired_settings = get_desired_settings_file(options.loadtest_name)
+    example_settings = \
+        get_example_settings_file(options.loadtest_name, options.environment)
```
rlucioni: The majority of our Python repos running pylint allow lines up to 120 chars.

rlucioni: https://github.com/edx/edx-lint

pavement.py
```diff
@@ -0,0 +1,81 @@
+from __future__ import print_function
+import os
+import os.path
+from paver.easy import task, sh, cmdopts
```
rlucioni: Nitpick: if pylint was running against this, it would encourage you to put this third-part import in its own block, below standard library imports.

pavement.py
pwnage101: paver only mentions optparse in the docs, so I assumed only that was supported.  It is possible to use paver's internal option parser but that's missing the ability to set defaults...so It would only be slightly more code to check for None + set default.

Perhaps a worthy sacrifice to get off optparse (but perhaps paver itself is transitively deprecated by depending on deprecated modules :P )

pavement.py
pwnage101: would have to see the python3 status of locustio/locust

rlucioni:
@pwnage101 done with my pass. I've left comments inline. Not to belabor the point, but note how much cleaner the [Invoke implementation](https://github.com/edx/edx-load-tests/commit/abb560e564af5d246480825bd37f3a4d8ccd701b) is.

loadtests/course_discovery/locustfile.py
pwnage101: the offending lines are around https://github.com/locustio/locust/blob/master/locust/main.py#L293

I have to run so I can't elaborate further, but I remember stepping through and noticing that the project root was not in sys.path at the time of loading the locustfile module.

pavement.py
rlucioni: @pwnage101 supported since July: https://github.com/locustio/locust/pull/363.

rlucioni: https://github.com/locustio/locust/blob/master/tox.ini#L2

rlucioni:
@pwnage101 your original Makefile PR included a target that could be used to run tests. I don't see a corresponding task here. Is that intentional?

loadtests/course_discovery/locustfile.py
rlucioni: @pwnage101 when you get back: how were you running locust? From the project root?

pavement.py
rlucioni: I'd like us to avoid introducing deprecated libraries when possible.

pwnage101:
> @pwnage101 your original Makefile PR included a target that could be used to run tests. I don't see a corresponding task here. Is that intentional?

Ah, I meant to write a comment about that.  It didn't make sense to me to abstract locust invokation because we already use a large subset of its features in the way described by locust --help.

However, it does occur to me that knowing to use ``-f loadtests/lms`` vs. ``-f lms`` is somewhat tribal knowledge (and subject to change).  For the rest of the args, it should be straightforward to passthrough args from paver to subprocesses.  Would something like this seem reasonable to you:

    $ paver loadtest --loadtest-name=lms [passthrough args]

loadtests/course_discovery/locustfile.py
pwnage101: yes, the root.  I'm beginning to remember what the exact sequence of events are:

1. user provides ``-f loadtests/lms`` on the command line
1. locust splits the locustfile path into "loadtests" and "lms"
1. locust adds "loadtests" to the path
1. locust calls ``__import__("lms")``, triggering our helper imports

That method of importing the locustfile precludes the possibility of referencing anything above the loadtests directory without adding it manually.

rlucioni:
@pwnage101 yes, seems reasonable.

loadtests/course_discovery/locustfile.py
rlucioni: I see. So it was working before because `edx-load-tests` (the project root) was being added to sys.path. Putting the tests in a `loadtests` directory causes `edx-load-tests/loadtests` to be added to sys.path instead, so you need to manually add `edx-load-tests` back to be able to use the helper imports.

I wonder why Locust imports the locustfile that way. If [this line](https://github.com/locustio/locust/blob/master/locust/main.py#L321) were to instead call `__import__('loadtests.course_discovery')`, this sys.path hacking wouldn't be necessary.

pwnage101:
This is still WIP, but pushing up my progress so @cpennington and @jzoldak can have some visibility.

jzoldak:
@pwnage101 I do think it would be easier to review if as you suggested you split off "reorg the repo" and "implement paver" into 2 separate PRs.

pwnage101:
splitting.

PR #74, opened 2016-09-28
implement a makefile again

pwnage101:
My makefile implementation in the past used a special GNU extension (`.SECONDEXPANSION`) which added complexity and was not portable to BSD make.  So I re-implemented it in pyinvoke.  Then paver.  Now I realize that what I had in Make was perfectly adequate, and I now know how to make it portable.  Here is my second try at using make.

This is part of PERF-357


pwnage101:
@rlucioni @tobz 

also FYI @cpennington

Makefile
```diff
@@ -1,4 +1,42 @@
...
+endif
+
+# figure out what all the loadtests are
+LOADTESTS := $(wildcard loadtests/*)  # get all files in the loadtests dir
+LOADTESTS := $(LOADTESTS:loadtests/%=%)  # get the loadtest name only
+LOADTESTS := $(filter-out __init__.py,$(LOADTESTS))  # filter out the __init__.py file
+
+help :
+	@echo 'see README.rst for usage details'
+	@echo '$(LOADTESTS)'
```
rlucioni: As it is, I don't find this particularly helpful. As someone approaching this for the first time, what am I supposed to do with these names?

Makefile
```diff
@@ -1,4 +1,42 @@
...
+	cp $< $@
+
+# this rule should be the only one that is required to run directly by users
+# and automation scripts, e.g. `make lms`
+$(LOADTESTS) : % : requirements settings_files/%.yml
+	@echo ""
+	@echo "HINT: Configure the $@ load test using the file:"
+	@echo "    settings_files/$@.yml"
+	@echo "HINT: Run the $@ load test with:"
+	@echo "    locust --host=<HOST> -f loadtests/$@"
```
rlucioni: I would expect this target to actually start locust for me, instead of spitting out a command I need to copy and paste.

Makefile
```diff
@@ -1,4 +1,42 @@
...
+LOADTESTS := $(wildcard loadtests/*)  # get all files in the loadtests dir
+LOADTESTS := $(LOADTESTS:loadtests/%=%)  # get the loadtest name only
+LOADTESTS := $(filter-out __init__.py,$(LOADTESTS))  # filter out the __init__.py file
+
+help :
+	@echo 'see README.rst for usage details'
+	@echo '$(LOADTESTS)'
+
+requirements :
+	pip install -r requirements.txt -q --exists-action w
```
rlucioni: I would remove the `-q` flag. Seeing the output about what's being installed is very helpful. Hiding it makes debugging more difficult than it needs to be.

Makefile
pwnage101: whoops, that line was from debugging

Makefile
pwnage101: As is, I agree (I'll will remove the flag).  In the future I was hoping to drop the seaborn dependency which appears to depend on numpy, and does not appear to be required for running any of the tests:

```
$ grep -RIi seaborn
requirements/base.txt:seaborn
util/Request Log Report.ipynb:    "import seaborn"
util/Request Log Report.ipynb:    "grid = seaborn.FacetGrid(\n",
util/Request Log Report.ipynb:    "grid = grid.map(seaborn.boxplot, \"bucket\", \"response_time\", whis=[5, 95])\n",
util/Request Log Report.ipynb:    "grid = seaborn.FacetGrid(\n",
util/Request Log Report.ipynb:    "grid = grid.map(seaborn.boxplot, \"bucket\", \"response_time\", showfliers=False, whis=[5, 95])\n",
```

The other requirements are not risky to install, and are fairly quick.

Makefile
pwnage101: If make were to invoke locust directly, it will likely exclude features such as distributed mode, no-web mode, etc., unless there was a make target for each of the different combinations (such a makefile would become difficult to maintain).  A makefile which excludes locust features is bad because a person running a load test would eventually have to understand how to run the tests using both methods (make + locust) anyway.  I'd be more interested in building out BenjiLee/locustauto to cover more use cases.

pwnage101:
Sorry to have this long a drawn out review process :( The paver implementation really ended up becoming prohibitively difficult to comprehend, and it was more like a sandalone python script rather than a set of interdependent paver tasks.  With pyinvoke being out of the question, I just decided to improve make and move on.

Makefile
rlucioni: @pwnage101 got it. A brief comment about why locust isn't being executed might be helpful to future maintainers with the same question.

rlucioni:
@pwnage101 that's totally fair, I just feel obligated to call things out.

Makefile
pwnage101: :+1:

Makefile
pwnage101: What about the preexisting `util/util_requirements.txt` file?  It looks like you committed the line containing seaborn, can you give more insight as to where it should go?

rlucioni: Sure, that requirements file sounds like a good fit. jupyter is also not required to run load tests, so you could consider moving it as well.

pwnage101:
addressed comments.  I moved runipy, seaborn, and jupyter to util/util_requirements.txt.

Makefile
```diff
@@ -1,4 +1,43 @@
...
+LOADTESTS := $(wildcard loadtests/*)  # get all files in the loadtests dir
+LOADTESTS := $(LOADTESTS:loadtests/%=%)  # get the loadtest name only
+LOADTESTS := $(filter-out __init__.py,$(LOADTESTS))  # filter out the __init__.py file
+
+help :
+	@echo 'see README.rst for usage details'
+
+requirements :
+	pip install -r requirements.txt --exists-action w
+
```
tobz: Can you add an additional target to install the optional dependencies, as well?  Does feel right to have a shortcut and then have to go the long way around for the rest.

pwnage101:
summary for latest commit:

* `make help` actually gives some help
* `make <test-name>` only installs pip requirements needed for running the load test, not pep8 or other development requirements.
* each load test gets a requirements.txt for specifying custom pip requirements.
* updated the README to suggest that each load test should have a requirements.txt
* deleted helpers/helper-requirements.txt (merged it into requirements/base.txt)
* the target for actually installing the settings file _really really_ doesn't overwrite existing settings files.

Makefile
pwnage101: next commit addresses this with a `make requirements` vs. `make util-requirements`.

pwnage101:
FYI @tobz

PR #75, opened 2016-09-30
implement the @interrupt_task (was @peek) decorator

pwnage101:
This decorator turns a normal locust task into one which greedily
interrupts (to interrupt from a task means to stop scheduling tasks from
the current TaskSet instance and change context to one TaskSet level up)

---

@robrap this is more or less complete, sans documentation.  it's late on friday, so I figured i'd just post it up and see what you think so far.

### DIFF
--- a/loadtests/lms/courseware_views.py
+++ b/loadtests/lms/courseware_views.py
@@ -1,4 +1,5 @@
 from locust import task
+from helpers.interrupt_after import interrupt_after
 
 from base import LmsTasks
 
@@ -23,6 +24,7 @@ class CoursewareViewsTasks(LmsTasks):
     """
 
     @task(50)
+    @interrupt_after
     def index(self):
         """
         Request a randomly-chosen top-level page in the course.
@@ -31,6 +33,7 @@ class CoursewareViewsTasks(LmsTasks):
         self.get(path, name='courseware:index')
 
     @task(33)
+    @interrupt_after
     def mktg_course_about(self):
         """
         Request the marketing about view (rendered as a button in the marketing site).
@@ -38,6 +41,7 @@ class CoursewareViewsTasks(LmsTasks):
         self.get('mktg-about', name='courseware:mktg_course_about')
 
     @task(10)
+    @interrupt_after
     def info(self):
         """
         Request the course info tab.
@@ -45,6 +49,7 @@ class CoursewareViewsTasks(LmsTasks):
         self.get('info', name='courseware:course_info')
 
     @task(4)
+    @interrupt_after
     def progress(self):
         """
         Request the progress tab.
@@ -52,6 +57,7 @@ class CoursewareViewsTasks(LmsTasks):
         self.get('progress', name='courseware:progress')
 
     @task(1)
+    @interrupt_after
     def about(self):
         """
         Request the LMS' internal about page for this course.
--- a/helpers/interrupt_after.py
+++ b/helpers/interrupt_after.py
@@ -1,0 +1,48 @@
+"""
+This file defines the interrupt_after decorator for locusts tasks.
+
+If your load test only has one TaskSet, ignore this utility.  Else, read on...
+
+Normally the locust client gets captured by a TaskSet class, preventing it from
+escaping to run tasks in sibling TaskSet classes unless it explicitly runs a
+task which calls the interrupt() function.  This interrupt_after decorator is
+simply a tool for calling the interrupt() function safely.
+
+If it is not detrimental to the load test results to have locust users hopping
+between TaskSets, you should consider using the interrupt_after decorator on
+all tasks of all TaskSets in your load test.  For small numbers of locust
+clients (especially when the test target is a load balancer) this will
+generally help keep the transaction distribution even.
+"""
+
+
+def is_root(task_set):
+    """
+    Return True if task_set is a root TaskSet instance, which means that it has
+    been called directly from the Locust class and so it has no parent.
+    """
+    # use `is` instead of `==` for reference equality
+    return task_set.parent is task_set.locust
+
+
+def interrupt_after(func):
+    """
+    Decorator for a locust task which causes it to be the last task to be
+    scheduled in the current TaskSet, unless the current TaskSet is the root
+    TaskSet---in the latter case this decorator is a no-op.
+
+    Use after the @task decorator:
+
+        @task(21)
+        @interrupt_after
+        def some_task(self): ...
+
+    """
+    def _interrupt_after(self, *args, **kwargs):
+        func(self, *args, **kwargs)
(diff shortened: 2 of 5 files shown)


helpers/peek.py
```diff
@@ -0,0 +1,31 @@
...
+LOGGER = logging.getLogger(__name__)
+
+def is_root(task_set):
+    """
+    Return True if task_set is a root TaskSet instance, which means that it has
+    been called directly from the Locust class and so it has no parent.
+    """
+    task_set.parent == task_set.locust
+
+def peek(func):
```
robrap: I guess I would have named this `interrupt`.  I don't understand the choice of the word `peek`?  Maybe you can explain here, and ultimately use it in the description if you decide to keep it.

helpers/peek.py
```diff
@@ -0,0 +1,31 @@
...
+        @task(21)
+        @peek
+        def some_task(self): ...
+
+    """
+    def _peek(self, *args, **kwargs):
+        func(self, *args, **kwargs)
+        # Before interrupting, make sure that there is a parent task set to
+        # escape to.  Otherwise locust will crash.
+        if not is_root(self):
```
robrap: Cool.

helpers/peek.py
```diff
@@ -0,0 +1,31 @@
...
+            self.interrupt()
+    return _peek
```
robrap: Nit: Maybe a line break between the def and this line?

loadtests/lms/courseware_views.py
```diff
@@ -23,6 +24,7 @@ class CoursewareViewsTasks(LmsTasks):
     """
 
     @task(50)
+    @peek
```
robrap: I don't know if this suggestion goes with or against best practices in Python, but if we think we will almost always want to use this the way we are coding, would it make sense to combine in a single `peek_task` or `interrupt_task`?  This is a question at this point, not a request.  I can also see an argument for keeping the decorators independent.  Thoughts?

robrap:
I'm glad this seems to be working out.  I would definitely send something out about this work when you have completed it.

robrap:
Sorry, I pushed the wrong comment button by mistake.  I didn't mean to close this.

helpers/peek.py
pwnage101: Whoops, I didn't explain.  The idea was that a task which "peeks" does not cause the user to linger, it just peeks.  I was just trying to use a name which intuitively describes its behavior, but wouldn't be conflated with the actual interrupt() function.

loadtests/lms/courseware_views.py
pwnage101: The primary reason for having task and peek separate decorators is that it allows the simultaneous use of peek and `self.tasks = {...}` to define the ratios centrally. e.g.

```python
@peek
task1():
    pass

@peek
task2():
    pass

self.tasks = { task1: 24, task2: 43 }
```

...or whatever the correct syntax would be.

helpers/peek.py
robrap: The term `peek` just isn't making sense to me.  Unfortunately, without knowing that the locust user gets stuck in a taskset until it sees an interrupt, no name will really make sense.  How about something like `exit_taskset`?  The comment could describe the normal behavior of locust and how this works.  Note that someone could also use it with a single task in a task set, the way locust documents `interrupt`, and the name of the decorator should still apply.

loadtests/lms/courseware_views.py
robrap: Got it.

helpers/peek.py
pwnage101: hm, what about `interrupt_after` or `exit_after` ?

robrap: Either of those are fine.  A comment that explains would be great.

pwnage101:
FWIW, I did look into implementing a _class_ decorator to automatically make every task in a TaskSet call interrupt,  but long story short the TaskSet class is itself derived from the TaskSetMeta metaclass which does some pre-processing on the tasks making it difficult to modify those tasks anymore.

pwnage101:
@robrap @tobz 

I left the removals of `return`s because they shouldn't be there anyway, but it is not related to this PR.

Users may set `LOCUST_TASK_SET` however they wish, and there are no consequences related to improper interrupt() calls (i tested with `LOCUST_TASK_SET` set to `LmsTest`, `ModuleRenderTasks`, etc.)

robrap:
👍 Glad you could get it to work for TaskSets run as root or as children.  Thanks.

tobz:
So, normally, as a child worker is assigned a task based on the task set/task ratios, they will permanently run that task?  And this gives them a chance to, once running the task once, go back to the "parent" task to potentially run another task?

pwnage101:
> So, normally, as a child worker is assigned a **taskset** based on the task set/task ratios, they will permanently run that **taskset**? And this gives them a chance to, once running **one task in the taskset**, go back to the "parent" **taskset** to potentially run another **taskset**?

@tobz FTFY

tobz:
Sure, ok, but what's the answer? :)

pwnage101:
Sorry, more complete answer is: no, what you wrote was incorrect.  TaskSets are sticky, not individual tasks.  The structure of the LMS tests currently look like:

* root taskset (e.g. LmsTest)
  * taskset A (e.g. ForumsTasks)
    * task 1 (e.g. search_topic)
    * task 2
  * taskset B
    * task 1
    * task 2

Despite the fact that the interrupt_task decorator decorates tasks, it's the stickiness of the taskset that it is actually concerned with.  I wasn't able to put together a clean/maintainable implementation of a class decorator, but that would be nicer for sure.

tobz:
So workers will get pinned to a taskset, like taskset A, but will migrate between tasks within the taskset they're pinned to?  Is that correct? (for how this PR works)

pwnage101:
That behavior which you describe is how locust normally works.  The decorator in this PR allows workers to break out of the taskset.

tobz:
Ah, I see.  So this will let us have, say, LmsTest, which has multiple things underneath it -- forums, courseware, etc -- and the global number of workers will just be able to freely roam from forums to courseware to whatever.

I just want to make sure I understand what it's being used for before I say "this seems like a good solution." :)

pwnage101:
> Ah, I see. So this will let us have, say, LmsTest, which has multiple things underneath it -- forums, courseware, etc -- and the global number of workers will just be able to freely roam from forums to courseware to whatever.

Exactly.  So when I run LmsTest two times, the transaction distribution will not change as much, especially when there is about 50 workers/host or less.  I believe this would have helped me collect cleaner data while testing small instances cleaner, or at least would have allowed me to run fewer/shorter tests.

> I just want to make sure I understand what it's being used for before I say "this seems like a good solution." :)

No problem, you're doing the right thing :)

PR #76, opened 2016-10-07
make sure secrets are redacted in command output

pwnage101:
Normally running a load test will print all the loaded settings to the console, even if they contain secrets.  In an attempt to prevent secrets getting logged in build artifacts or wherever the console log may end up, I simply filter the output to redact the secrets (in this PR).

This happens to also implement a make target to run pytest.   Not totally sure if the method I use is good or normal.

@cpennington we discussed making sure secrets don't get printed to stdout

@rlucioni I mostly modeled the testing structure around the make targets in course-catalog, so I figured i'd tag you too.


pwnage101:
Hm...I can't get the tests to successfully import anything in the unit tests in travis.  Locally I had to run `pip install -e .` but i'm not sure about the correct way to making this work with travis.

rlucioni:
I feel obligated to ask: is it really a problem to have secrets from load test environments leaked to the console? What could a malicious actor with that information do?

rlucioni:
@pwnage101 why not just do away with logging config to the console by default? You could leave the functionality in, but make it optional for debugging purposes. That seems like a much more thorough solution to the problem at hand.

pwnage101:
I think it's important to somehow report the input settings at all.  For the same reason that it's important to know the client count or the run time, input settings should be recorded.  Otherwise they should be hard-coded in the code and made non-optional, IMO.

Printing input settings to the console is currently the easiest way to report them.  In the future I can see load test runs producing build artifacts which are browsable via the Jenkins web UI, for example, so this PR is somewhat proactive.

helpers/settings.py
```diff
@@ -1,29 +1,65 @@
...
+
+"""
 import os
 import yaml
 import logging
+import copy
 from pkg_resources import resource_filename
 from pprint import pformat
 
+REDACTED_KEYWORDS = ['password', 'secret']
```
pwnage101: I agree this needs attention; a weak filter can be dangerous because it gives a false sense of security.  In the case of nested dicts, that already has a test case.  I'm not sure what you mean by "flat client ID and key/secret" though.

pwnage101:
one more note:

> What could a malicious actor with that information do?

For instance, if a malicious actor gets a hold of the basic auth defaults, their blast radius is quite limited but that still means we would need to change the defaults again, update internal docs, broadcast the change, etc.

helpers/settings.py
rlucioni: What I meant was that I might structure my config like this:
```yaml
oauth:
  provider_url: https://example.com
  client_id: foo
  client_key: bar
```
Or like this:
```yaml
provider_url: https://example.com
client_id: foo
client_key: bar
```
In either case, all of this config would be leaked by the existing filter. You hit the nail on the head as far as the current approach goes: it instills a false sense of security. As it stands, if we want to be confident that secrets won't be leaked to the console, they shouldn't be printed in the first place. If outputting this config as a build artifact is somehow more secure, we should be doing that instead of what is ultimately a half-measure.

pwnage101:
Hm, it sounds like this isn't going anywhere.  I'd like to float another
idea: to specify secrets in a separate file or a separate YAML document.

In the latter case, the norm could be that values outside of the first
document are secrets. E.g.

```yaml
---
username: please-set-me
---
# secrets
password: please-set-me
```

Or with oauth dict settings:

```yaml
---
oauth:
  id: please-set-me
---
# secrets
oauth:
  secret: please-set-me
```

Then they could be referenced separately, e.g. `settings.data` vs.
`settings.secrets`.

Thoughts?

rlucioni:
@pwnage101 sorry about the delay, I was out of the country. A separate document for secrets seems reasonable and more complete to me.

pwnage101:
@rlucioni Added another squash commit to implement the idea in my previous comment.  Tests are a little more end-to-end now.  I have yet to manually test this, will update after I do that.

rlucioni:
Nice, thanks for the heads-up @pwnage101.

pwnage101:
I tested it, and it seems to work.  For that I used the lms loadtest and the BASIC_AUTH_* secrets.  Also noteworthy that the travis checks are passing because this PR introduces unit tests for this repository.

I dropped some commits related to the old implementation, so all the diffs should be relevant to the current implementaiton.  Basically the first couple commits set up testing infrastructure, then starting with "squash me 6" they implement the secrets segregation stuff.

@rlucioni

helpers/settings.py
```diff
@@ -42,14 +70,35 @@ def init(test_module_full_name, required=[]):
...
+        LOG.info('secrets loaded from the settings file')
+    else:
+        LOG.info('no secrets were specified in the settings file')
+    LOG.info('loaded the following public settings:\n{}'.format(
+        pformat(data),
+    ))
 
-    # check that the required keys are present
+    # check that the required settings are present
     for key in required:
```
rlucioni: I'd be consistent with the below and make this something like `required_data`.

requirements/dev.txt
```diff
@@ -2,3 +2,5 @@
 -r base.txt
 
 pep8==1.6.2
+pytest==3.0.3
+mock==2.0.0
```
rlucioni: Minor: keep these alphabetized.

settings_files/course_discovery.yml.example
```diff
@@ -13,3 +12,10 @@ programs:
     - foo
     - bar
     - baz
+
+---
+# secrets below
+
+oauth:
```
rlucioni: I'd suggest moving the entire `oauth` hash into the secrets section. Splitting it like this strikes me as unnecessarily complicated.

settings_files/ecommerce.yml.example
```diff
@@ -14,3 +14,11 @@ ecommerce:
 jwt:
   secret_key: set-me-please
   issuer: https://127.0.0.1:8000/oauth2
+
+---
+# secrets below
+
+ecommerce:
+  cybersource_secret_key: set-me-please
+jwt:
```
rlucioni: Move the entire `jwt` hash into secrets. Don't leave `cybersource_secret_key` in the data section, since we expect it in secrets.

tests/test_settings.py
```diff
@@ -0,0 +1,51 @@
...
+        settings.secrets = None
+
+    @patch('helpers.settings.resource_filename')
+    def test_load_settings(self, mock_resource_filename):
+        mock_resource_filename.return_value = 'tests/foo.yml'
+        settings.init('loadtests.foo.locustfile', ['REQUIRED_KEY'], ['REQUIRED_SECRET_KEY'])
+        mock_resource_filename.assert_called_once_with('settings_files', 'foo.yml')
+        with open('tests/foo.yml') as foo:
+            foo_data, foo_secrets = yaml.load_all(foo)
+        assert settings.data == foo_data
```
rlucioni: I see you're using `unittest` here, but not the `self.assert*` methods. I know `pytest` provides information about failing asserts (which I prefer to `unittest`'s methods), but is it conventional to mix `unittest` and naked asserts like this? It seems a little odd.

pwnage101: Oops, i thought unittest was more like a helper library rather than something that overlapped with pytest.  I may tweak this to be pure pytest.

helpers/settings.py
pwnage101: makes sense

pwnage101:
@rlucioni one more pass please!

.travis.yml
```diff
@@ -10,9 +10,11 @@ cache:
   directories:
     - $HOME/.cache/pip
 install:
-    - pip install --upgrade pip wheel pep8
+    - pip install --upgrade pip wheel
+    - make requirements
+    - pip install -e .
```
rlucioni: This is a weird way of installing requirements, especially since this repo isn't really meant to be used as a pip-installable package. Why not just `pip install -r requirements/dev.txt`?

Makefile
```diff
@@ -39,6 +42,17 @@ help :
 	@echo '        sandbox-specific settings are present under'
 	@echo '        settings_files/lms.sandbox-example.'
 
+clean :
+	find . -name '*.pyc' -delete
+
+test : clean
+	py.test tests
```
rlucioni: Run these in parallel al-la https://github.com/edx/tubular/blob/master/Makefile#L7.

pwnage101:
```
$ time py.test tests
[...]
py.test tests  0.49s user 0.10s system 98% cpu 0.593 total

$ time py.test -n auto tests
[...]
py.test -n auto tests  1.21s user 0.24s system 90% cpu 1.606 total
```

parallel scheduling overhead is greater :P

I'll throw it in anyway since hopefully we will write more tests...

rlucioni: edx-platform is a good example of what happens when you don't run your tests in parallel from the start.

.travis.yml
pwnage101: I think it needs itself installed because the tests won't run otherwise:

```
$ py.test tests
[...]
================================================================================== ERRORS ===================================================================================
__________________________________________________________________ ERROR collecting tests/test_settings.py __________________________________________________________________
ImportError while importing test module '/home/tsankey/tmp/edx-load-tests/edx-load-tests/tests/test_settings.py'.
Original error message:
'No module named helpers'
Make sure your test modules/packages have valid Python names.
```

Are you suggesting to put `pip install -e .` inside `requirements/dev.txt`?

rlucioni: No, I just thought you were using `pip install -e .` to install requirements. Requiring that the package itself be installed so that you can run tests is pretty unconventional. It'd be great if we could avoid the need for this.

pwnage101: Ah, okay I think I figured out the problem:

I didn't create a `tests/__init__.py` file because the pytest "Good Integration Practices" page [0] basically says to univerally avoid it, but also makes the assumption that the project in question is designed to be a pip installable package.  In my case, creating that file fixes the problem.

[0] http://doc.pytest.org/en/latest/goodpractices.html

rlucioni: Nice find.

rlucioni:
@pwnage101 you didn't merge your commits! 😭

pwnage101:
literally one second after clicking merge i tried to cancel, and the button froze.  I immediately rewrote history...gonna squash them now...

PR #77, opened 2016-11-15
Add new proctoring load tests.

dianakhuang:
@robrap @pwnage101 

Here's the current version of the proctoring tests. I didn't tweak the numbers yet, since I figured we'd modify those once we have better data.

### DIFF
--- a/loadtests/lms/course_data/__init__.py
+++ b/loadtests/lms/course_data/__init__.py
@@ -66,6 +66,15 @@ class CourseData(dict):
         """
         return random.choice(self['html_usage_ids'])
 
+    @property
+    def exam_id(self):
+        """
+        Randomly select an exam id from all known special exams in this course.
+
+        Special exams are either timed exams or proctored exams.
+        """
+        return random.choice(self['special_exam_ids'])
+
 
 # data extracted from edX/DemoX/Demo_Course as of Feb. 2015
 demo_course = CourseData(
--- a/loadtests/lms/locustfile.py
+++ b/loadtests/lms/locustfile.py
@@ -12,6 +12,7 @@ from locust import HttpLocust
 from courseware_views import CoursewareViewsTasks
 from forums import ForumsTasks, SeedForumsTasks
 from base import LmsTasks
+from proctoring import ProctoredExamTasks
 from module_render import ModuleRenderTasks
 from wiki_views import WikiViewTask
 
@@ -99,7 +100,8 @@ class LmsTest(LmsTasks):
     tasks = {
         CoursewareViewsTasks: 9,
         ForumsTasks: 2,
-        ModuleRenderTasks: 27,
+        ModuleRenderTasks: 27 * float(settings.data.get('MODULE_RENDER_MODIFIER', 1)),
+        ProctoredExamTasks: 1 * float(settings.data.get('PROCTORED_EXAM_MODIFIER', 1)),
     }
 
 
--- a/loadtests/lms/proctoring.py
+++ b/loadtests/lms/proctoring.py
@@ -1,0 +1,44 @@
+import json
+from locust import task
+
+from base import LmsTasks
+
+
+class ProctoredExamTasks(LmsTasks):
+    """
+    Tasks representing the accessing of the proctored exam endpoints.
+
+    These endpoints are used by both timed exams and proctored exams to sync the
+    browser state with the server state. Both proctored exams and timed exams
+    share their code.
+    """
+
+    attempt_id = None
+    attempt_api_path = '/api/edx_proctoring/v1/proctored_exam/attempt'
+
+    def on_start(self):
+        super(ProctoredExamTasks, self).on_start()
+        self.create_attempt()
+
+    def create_attempt(self):
+        """
+        Create an attempt for the current user.
+        """
+        data = {'exam_id': self.course_data.exam_id, 'start_clock': True}
+        response = self.client.post(self.attempt_api_path, data=data, headers=self.post_headers)
+        response_data = json.loads(response.text)
+        self.attempt_id = response_data.get('exam_attempt_id')
+
+    @task(1)
+    def exam_poll_attempt(self):
+        """
+        Retrieves the status of an existing exam attempt.
+
+        This mimics the polling that the banner of a timed exam or proctored exam performs.
+        """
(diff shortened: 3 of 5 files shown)


loadtests/lms/course_data/__init__.py
```diff
@@ -66,6 +66,11 @@ def html_usage_id(self):
         """
         return random.choice(self['html_usage_ids'])
 
+    @property
+    def exam_id(self):
+        """ Randomly select an exam id from all known special exams in this course. """
```
robrap: Nit: Move text to its own line.

loadtests/lms/proctoring.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def on_start(self):
+        super(ProctoredExamTasks, self).on_start()
+        self.create_attempt()
+
+    def create_attempt(self):
+        """ Create an attempt for the current user. """
+        data = {'exam_id': self.course_data.exam_id, 'start_clock': True}
+        response = self.client.post(self.url, data=data, headers=self.post_headers)
+        print response
```
robrap: was the print meant to be temporary?

loadtests/lms/proctoring.py
```diff
@@ -0,0 +1,35 @@
...
+
+    attempt_id = None
+    url = '/api/edx_proctoring/v1/proctored_exam/attempt'
+
+    def on_start(self):
+        super(ProctoredExamTasks, self).on_start()
+        self.create_attempt()
+
+    def create_attempt(self):
+        """ Create an attempt for the current user. """
```
robrap: Nit: Commit format (and through file).  Add text on its own line.

loadtests/lms/proctoring.py
```diff
@@ -0,0 +1,38 @@
+import json
+from locust import task
+
+from base import LmsTasks
+
+
+class ProctoredExamTasks(LmsTasks):
+    """
+    Tasks representing the accessing of the proctored exam endpoints.
```
robrap: Could you help clarify use of proctored exams vs timed exams in the comments?

loadtests/lms/proctoring.py
```diff
@@ -0,0 +1,38 @@
...
+    """
+
+    attempt_id = None
+    url = '/api/edx_proctoring/v1/proctored_exam/attempt'
```
pwnage101: For better self-documentation, I would like "url" to have a more accurate variable name such as "api_path", or "attempt_path"

loadtests/lms/course_data/__init__.py
```diff
@@ -66,6 +66,13 @@ def html_usage_id(self):
         """
         return random.choice(self['html_usage_ids'])
 
+    @property
+    def exam_id(self):
+        """
+        Randomly select an exam id from all known special exams in this course.
+        """
+        return random.choice(self['special_exam_ids'])
```
robrap: I prefer the pattern in on_start of the forums tests that does a one time query of special_exam_ids from the course, rather than relying on configuration.  It simply makes the test easier to run.  I'm not sure how easy or difficult this would be.  

Also, do we need to clarify here and near it definition of `special_exam_ids` exactly what type of exams will work?

loadtests/lms/proctoring.py
```diff
@@ -0,0 +1,38 @@
...
+        """
+        Create an attempt for the current user.
+        """
+        data = {'exam_id': self.course_data.exam_id, 'start_clock': True}
+        response = self.client.post(self.url, data=data, headers=self.post_headers)
+        response_data = json.loads(response.text)
+        self.attempt_id = response_data.get('exam_attempt_id')
+
+    @task(1)
+    def attempt(self):
```
robrap: Is this the polling?  Maybe rename to `attempt_status`, or `poll_attempt`, or something of that nature?  Also, maybe mention a bit more about the polling in the comment below.

Are we going to load an exam problem and/or making an exam decision or make some other call to replicate the test taking?  I have a feeling it might make it simpler to tune the overall test for regular production vs a large proctoring test.

loadtests/lms/locustfile.py
```diff
@@ -100,6 +101,7 @@ class LmsTest(LmsTasks):
         CoursewareViewsTasks: 9,
         ForumsTasks: 2,
         ModuleRenderTasks: 27,
+        ProctoredExamTasks: 1,
```
robrap: How do we plan on differentiating between normal timed exam load added into the LMS tests and proctored exam heavy load testing?

loadtests/lms/course_data/__init__.py
dianakhuang: I don't think there's an endpoint for getting all the ids from a course that currently exists. Are you recommending that I add one?

loadtests/lms/proctoring.py
dianakhuang: Those interactions with exam problems and test-taking should be covered by the existing `module_render` tests. Interacting with those problems during an exam is no different from interacting with them outside an exam. The only real difference is that there is polling occurring at the same time.

loadtests/lms/locustfile.py
pwnage101: assuming that a proctored exam can be approximated using timed exam task ratios with a higher poll rate, then i think the poll ratio within the ProctoredExamTasks task set needs to be configurable somehow.  I do think that it might be out of scope for this PR, however.

dianakhuang:
@robrap @pwnage101 I believe I've addressed your comments.

settings_files/lms.yml.example
```diff
@@ -17,6 +17,10 @@ LOCUST_TASK_SET: LmsTest
 #LARGE_TOPIC_ID:
 #LARGE_THREAD_ID:
 
+# Used to determine how 'active' users are in timed and proctored exams.
+# This is the multiplying factor for how often proctoring-specific tasks happen:
+#PROCTORED_EXAM_MODIFIER: 10
+
```
pwnage101: Please either specify the default value or make PROCTORED_EXAM_MODIFIER a required setting.  Part of my plan to make load tests easier to run is to make settings as transparent as possible.

pwnage101:
Sorry, I meant to specify the default in the above comment, rather than uncommenting the setting.

robrap:
👍

PR #79, opened 2016-11-29
locust task ratios can only be integers

pwnage101:
Before this commit, the LMS loadtest would just crash on my machine.
Locust does not appear to tolerate float ratios.

---

@dianakhuang


dianakhuang:
👍

PR #80, opened 2016-11-30
update module_render task distribution

pwnage101:
also, add helper script: lms_module_render_tx_distribution.py

### DIFF
--- a/loadtests/lms/module_render.py
+++ b/loadtests/lms/module_render.py
@@ -11,22 +11,224 @@ class ModuleRenderTasks(LmsTasks):
     """
     Tests callback handlers implemented in lms.djangoapps.courseware.module_render
 
-    NewRelic reports virtually all of this traffic under the single function call,
-    module_render.handle_xblock_callback.   The following list of top endpoints
-    comes from a rough analysis of access logs for the 7 days prior to 2015-02-25.
+    See util/lms_module_render_tx_distribution.sh for instructions on
+    generating the data below.
 
-    xmodule_handler/problem_get      5157055  21.22%
-    xmodule_handler/save_user_state  5144295  21.17%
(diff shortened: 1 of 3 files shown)


pwnage101:
@dianakhuang @robrap

loadtests/lms/module_render.py
```diff
@@ -11,22 +11,224 @@ class ModuleRenderTasks(LmsTasks):
...
-    goto_position                    538610   2.22%
-    xmodule_handler/problem_show     466705   1.92%
-    save_user_state                  447787   1.84%
-    handle_grade                     291978   1.20%
-    problem_get                      244093   1.00%
-    xmodule_handler/problem_save     226407   0.93%
+    See util/lms_module_render_tx_distribution.sh for instructions on
+    generating the data below.
+
+    module_render endpoint distribution (last 7 days as of 2015-02-25):
```
pwnage101: What originally prompted this was that problem_get went from the most called task to the least called task.  We aren't necessarily concerned with the minute differences between courses, but more about these broad platform changes.

dianakhuang:
👍  from me.

loadtests/lms/module_render.py
pwnage101: > Do we need to take into account peak load on heavier endpoints?

I think this would lead to higher accuracy but I don't how much higher.  Anyway I'll share my idea:

If we pulled out ratio definitions into a yaml file (e.g. ratios.yml) defining complete sets of task ratios, we could define different ratio sets for different endpoint distributions.  For example, assuming different courses each have a different endpoint distributions, we could have ratios_demo_course.yml and ratios_supply_chain_course.yml.

Out of scope for this PR.

robrap:
👍

PR #85, opened 2016-12-29
Update requirements for locust version

davec-edx:
ECOM-6711
@edx/ecommerce

Updated requirements to use locust 0.7.5 instead of 0.7.3.


MatthewPiatetsky:
👍

PR #86, opened 2017-01-03
[lms] remove all interrupt_after calls

pwnage101:
interrupt()ing after every request causes locust to re-run on_start(),
causing the number of auto_auth and enroll transactions to shoot up.
This is worse than not using the decorator so we just remove it for
now.

This commit also fixes another bug with the decorator: force locust to
obey the wait/delay period after calling interrupt().

---
 @dianakhuang


dianakhuang:
:+1:

PR #88, opened 2017-01-12
fix settings regression with non-existant secrets document

pwnage101:
Add another test for settings files which have two documents, but the second document (secrets) is empty.  The yaml library produces `None` on empty documents instead of an empty dict `{}`, so capture this case and force settings.secrets = `{}`.

e.g. before this PR, the standard `settings_files/lms.yml.example` stops working because it contains an empty secrets document.


pwnage101:
@rlucioni

PR #89, opened 2017-01-17
Pin gevent version

efischer19:
Aha!

It turns out that, in a fresh virtualenv, the newly-updated version 0.7.5 of locustio will ask for gevent==1.1.1 to be installed. This version of gevent breaks something related to socket I/O for our use case (throws a low-level segfault). I was able to sidestep this issue by pinning gevent to a specific version (one that was already globally installed on my machine).

@pwnage101, any objections to this?

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -2,6 +2,7 @@
 # should import this file.
 
 # load testing software
+gevent==1.0.2
 locustio==0.7.5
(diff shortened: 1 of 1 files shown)


pwnage101:
@efischer19 sorry for the delay, have you tested this on ubuntu?

pwnage101:
for my next test runs i'll give this a try, then merge if it works

PR #91, opened 2017-01-30
[lms] drive load against multiple courses

pwnage101:
This allows multiple courses to be specified in settings like such:

```yaml
courses:
    edX/DemoX/Demo_Course:
        course_data: demo_course
        ratio: 5
    course-v1:edX+DemoX+Demo_Course:
        course_data: demo_course_split
        ratio: 3
...
```

Note that the locust summary of response times broken out by request type is the aggregate of all courses.  A possible PR for the future would additionally break out locust results by course.

@ormsbee

### DIFF
--- a/helpers/util.py
+++ b/helpers/util.py
@@ -1,0 +1,39 @@
+"""
+Generally helpful utilities for edx load testing.
+"""
+from random import random
+
+
+def choice_with_distribution(element_ratio_pairs):
+    """
+    Randomly choose an element with probabilities given by ratios.
+
+    This utility function takes a mapping of elements to ratios and chooses an
+    element with probability proportional to the corresponding ratio.  The
+    ratios do not need to add up to 1.0.
+
+    Args:
+        element_ratio_pairs (list): List of two-tuples containing the element
+            and its ratio (int or float).  E.g. [('hello', 12), ('world', 20)]
+
+    Returns:
+        An element (the first item in one of the given tuples)
+
+    Raises:
+        ValueError: If the input list is empty.
+        RuntimeError: If there is a bug in this function which causes no
+            element to get chosen.
+    """
+    if not element_ratio_pairs:
+        raise ValueError('element_ratio_pairs is empty.')
+    ratios_sum = sum(pair[1] for pair in element_ratio_pairs)
+    random_variable = random()  # possible values are [0.0, 1.0)
+    # The element is chosen based on iteratively decreasing the random variable
+    # by the normalized ratio of each element until it becomes negative.
+    for elem, ratio in element_ratio_pairs:
+        ratio_normalized = float(ratio) / ratios_sum
+        random_variable -= ratio_normalized
+        if random_variable < 0:
+            return elem
+    raise RuntimeError(
+        "Exited element selection loop without making a selection.")
--- a/loadtests/lms/base.py
+++ b/loadtests/lms/base.py
@@ -4,12 +4,14 @@
 import logging
 import os
 import sys
+import random
 
 from lazy import lazy
 from opaque_keys.edx.keys import CourseKey
 
 from helpers.auto_auth_tasks import AutoAuthTasks
 from helpers import settings
+from helpers import util
 
 import course_data
 
@@ -23,8 +25,17 @@ class EdxAppTasks(AutoAuthTasks):
     def course_id(self):
         """
         The complete id of the course we're configured to test with.
+
+        This function is evaluated once per locust client and its return value
+        is cached on a per-client basis (due to the @lazy decorator).  It
+        randomly selects a course from the "courses" dict specified in the
+        settings file.  The "ratio" keys of every course are used to construct
+        a probability distribution.
         """
-        return settings.data['COURSE_ID']
+        courses = settings.data['courses']
+        course_ratio_pairs = \
+            [(cid, cdata['ratio']) for cid, cdata in courses.iteritems()]
+        return util.choice_with_distribution(course_ratio_pairs)
 
     @lazy
     def course_key(self):
@@ -59,8 +70,21 @@ class EdxAppTasks(AutoAuthTasks):
         """
         Accessor for the CourseData instance we're configured to test with.
         """
-        course_data_name = settings.data['COURSE_DATA']
+        course_data_name = self._get_course_setting('course_data')
         return getattr(course_data, course_data_name)
+
+    def _get_course_setting(self, setting):
+        """
+        Accessor for a setting specific to the current course.
+
+        Args:
+            setting (str): name of course-specific setting
+
+        Raises:
+            KeyError: If the setting was not specified for this locust client's
+                selected course_id.
+        """
+        return settings.data['courses'][self.course_id][setting]
 
     @property
     def post_headers(self):
--- a/loadtests/lms/forums.py
+++ b/loadtests/lms/forums.py
@@ -72,7 +72,10 @@ class BaseForumsTasks(LmsTasks):
         if BaseForumsTasks._large_thread:
             return BaseForumsTasks._large_thread[0]
         else:
-            return settings.data.get('LARGE_TOPIC_ID')
+            try:
+                return self._get_course_setting('large_topic_id')
+            except KeyError:
+                return None
 
     @lazy
     def large_thread_id(self):
@@ -82,7 +85,10 @@ class BaseForumsTasks(LmsTasks):
         if BaseForumsTasks._large_thread:
             return BaseForumsTasks._large_thread[1]
         else:
-            return settings.data.get('LARGE_THREAD_ID')
+            try:
+                return self._get_course_setting('large_thread_id')
+            except KeyError:
+                return None
 
     def create_thread(self, topic_id, name):
         """
--- a/loadtests/lms/locustfile.py
+++ b/loadtests/lms/locustfile.py
@@ -19,8 +19,7 @@ from tracking import TrackingTasks
 
 from helpers import settings
 settings.init(__name__, required_data=[
-    'COURSE_ID',
-    'COURSE_DATA',
+    'courses',
     'LOCUST_TASK_SET',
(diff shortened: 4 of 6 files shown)


pwnage101:
@dianakhuang can you also take a look

loadtests/lms/base.py
```diff
@@ -24,7 +26,12 @@ def course_id(self):
         """
         The complete id of the course we're configured to test with.
         """
-        return settings.data['COURSE_ID']
+        courses = settings.data['courses']
+        # Create two lists, a list of course IDs and a list of course ratios,
+        # such that both lists are in the same order.
+        course_ids, course_ratios = \
+            zip(*((cid, cdata['ratio']) for cid, cdata in courses.iteritems()))
+        return util.choice_with_distribution(course_ids, course_ratios)
```
ormsbee: This method has a `@lazy` decorator, which I assume means it'll be cached? Won't that break the random distribution?

loadtests/lms/base.py
```diff
@@ -24,7 +26,12 @@ def course_id(self):
         """
         The complete id of the course we're configured to test with.
         """
-        return settings.data['COURSE_ID']
+        courses = settings.data['courses']
+        # Create two lists, a list of course IDs and a list of course ratios,
+        # such that both lists are in the same order.
+        course_ids, course_ratios = \
+            zip(*((cid, cdata['ratio']) for cid, cdata in courses.iteritems()))
```
ormsbee: This is a pretty dense line to unpack. Is this the clearest data structure to have? As opposed to, say, a `dict` mapping `course_id` to weights, e.g.:

```python
course_weights = { course_id: data['ratio'] for course_id, data in courses.items() }
```

loadtests/lms/base.py
ormsbee: Also, how often is this called? The caching decorator makes me wonder if this may become a performance bottleneck.

loadtests/lms/forums.py
```diff
@@ -72,7 +72,10 @@ def large_topic_id(self):
         if BaseForumsTasks._large_thread:
             return BaseForumsTasks._large_thread[0]
         else:
-            return settings.data.get('LARGE_TOPIC_ID')
+            try:
+                return self._get_course_setting('large_topic_id')
+            except KeyError:
+                return None
```
ormsbee: Why is it ok to return `None` here?

loadtests/lms/forums.py
```diff
@@ -82,7 +85,10 @@ def large_thread_id(self):
         if BaseForumsTasks._large_thread:
             return BaseForumsTasks._large_thread[1]
         else:
-            return settings.data.get('LARGE_THREAD_ID')
+            try:
+                return self._get_course_setting('large_thread_id')
+            except KeyError:
+                return None
```
ormsbee: Same question regarding `None`

helpers/util.py
```diff
@@ -0,0 +1,23 @@
...
+from random import random
+
+
+def choice_with_distribution(elements, ratios):
+    """
+    Randomly choose an element according to the given ratios.  The ratios are
+    proportional to a probability distribution, but need not add up to 1.0.
+    """
+    if len(elements) == 0:
+        raise ValueError('elements list is empty.')
```
ormsbee: `if not elements:` is more idiomatic

helpers/util.py
```diff
@@ -0,0 +1,23 @@
...
+Generally helpful utilities for edx load testing.
+"""
+from random import random
+
+
+def choice_with_distribution(elements, ratios):
+    """
+    Randomly choose an element according to the given ratios.  The ratios are
+    proportional to a probability distribution, but need not add up to 1.0.
+    """
```
ormsbee: Please specify in the docstring the data structures being passed in and returned.

helpers/util.py
```diff
@@ -0,0 +1,23 @@
...
+    ratios_normalized = (float(r) / sum(ratios) for r in ratios)
+    random_variable = random()  # possible values are [0.0, 1.0)
+    # The element is chosen based on iteratively decreasing the random variable
+    # by the normalized ratio of each element until it becomes negative.
+    for elem, rn in zip(elements, ratios_normalized):
+        random_variable -= rn
+        if random_variable < 0:
+            return elem
+    raise RuntimeError(
+        "Exited element selection loop without making a selection.")
```
ormsbee: Does this operation get simpler if we constrain it to integers and use something like `random.choice`? I ask mostly because the `RuntimeError` at the end there seems to imply that we're not confident if it'll return correctly in all cases?

pwnage101: What exactly are you suggesting that we constrain to integers?  The random.choice function assumes a uniform probability distribution, but I want to supply an alternative distribution (the ratios list).

loadtests/lms/base.py
pwnage101: Indeed, I could create a dict mapping and pass that into choice_with_distribution.

loadtests/lms/base.py
pwnage101: This class is instantiated once for each locust client.  Each locust client is forced to evaluate this in the very beginning, and then gets stuck with it until the end of the test.  Without the lazy decorators in this class, it would get called on every request.

loadtests/lms/forums.py
pwnage101: I selected `None` to avoid change in behavior.  `dict.get` would also return `None` if the key did not exist.

loadtests/lms/base.py
pwnage101: I now remember why `choice_with_distribution` needs to take two separate lists.  There's no guarantee that the elements are hashable (course_id in this case, which is of type string), and there's a possibility that the ratios (integers) could be the same across multiple elements causing a collision.  Neither the elements, nor the ratios can be the key of a dict mapping.

I willl opt to unpack this zip code into something more readable, but maintain the `choice_with_distribution` interface.

loadtests/lms/base.py
ormsbee: Got it, thanks.

loadtests/lms/base.py
ormsbee: Sorry, I'm not clear on why `course_id` would ever not be hashable. Do you mean there could be multiple entries for the same `course_id`? Strings are definitely hashable, as are all OpaqueKey types.

loadtests/lms/base.py
```diff
@@ -59,9 +66,18 @@ def course_data(self):
...
         Accessor for the CourseData instance we're configured to test with.
         """
-        course_data_name = settings.data['COURSE_DATA']
+        course_data_name = self._get_course_setting('course_data')
         return getattr(course_data, course_data_name)
 
+    def _get_course_setting(self, setting):
+        """
+        Accessor for a setting specific to the current course.  Missing
+        settings are assumed to be fatal.
```
ormsbee: Is it really fatal? The only places that call this method diaper it over by catching the `KeyError` and return `None`.

loadtests/lms/forums.py
ormsbee: BTW, is there a way to tell which settings are mandatory-and-we-will-explode-as-misconfigured, and which ones are optional?

loadtests/lms/base.py
pwnage101: I'm saying that somebody may try/want to use `choice_with_distribution` in `helpers/util.py` with unhashable objects.  It is true that I am using it with hashable objects, but my secondary goal is to provide this possibly useful function to anybody who needs it.

loadtests/lms/forums.py
pwnage101: Normally we codify required settings during initialization of the settings, e.g.:

```python
from helpers import settings
settings.init(__name__, required_data=[
    'courses',
    'LOCUST_TASK_SET',
    'LOCUST_MIN_WAIT',
    'LOCUST_MAX_WAIT',
])
```

This gets evaluated so early in the process that it's sort of pre-runtime.  If the required setting exists but the value happens to be a dict itself, it's up to the runtime code (e.g. code in `loadtests/lms/base.py`) to determine whether missing settings are fatal.  I wrote the settings code before considering the use case of hierarchical settings.

loadtests/lms/base.py
ormsbee: Ah, I see.

FWIW, the reason why I'm nitpicking about this is because I've found it to be more bug prone to be passing in two variables that have to be in sync with each other in some way (two matching lists), than one data structure. For instance, `choice_with_distribution` currently won't fail if the arguments are of different lengths (`zip` will silently stop with the shortest sequence), even though it's an invalid input state. One approach is to check for all the invalid states up front, but it's best if we can make those invalid states less likely to occur in the first place.

So if a dict isn't safe because of the hashable arg issue, then a list of `(arg, weight)` tuples might be more appropriate. But I want to urge you to embrace functional limitations if it simplifies your code. There's a pretty good chance that `choice_with_distribution` is never going to need to handle unhashable items.

helpers/util.py
ormsbee: The weight ratios for instance. If they were integers, you could say something this:

```python
from itertools import chain, repeat
import random

def choice_with_distribution(elements_to_weights):
    # Denormalize the input, so {'a':2, 'b':3} becomes ['a', 'a', 'b', 'b', 'b']
    choices = chain.from_iterable(
        repeat(element, weight)
        for element, weight in elements_to_weights.items()
    )
    return random.choice(list(choices))
```

Not super efficient, but it requires no actual thinking about correctness.

This is not a required change. I realize that the algorithm you're using isn't rocket science. But I wanted to at least float the suggestion.

loadtests/lms/base.py
pwnage101: Fatal is really the wrong word to use here.  I will fix it.

pwnage101:
> Just to confirm that I understand this code correctly, each hatched locust will be associated with a specific course/course_id and will only interact with that course?

@dianakhuang yes, exactly.

PR #92, opened 2017-02-02
Add existing comments service and notes load tests.

dianakhuang:
Old tests. Need to be updated and moved over to the public repo.

FYI @maxrothman


pwnage101:
Can you also add README files for each of these loadtests to describe that they should be integrated with the settings files.  Also I think both tests need a `__init__.py` file.

dianakhuang:
@pwnage101 i believe these aren't necessary anymore for our load testing, so I think we can close this PR. I'll do that now, and if we need these again, I'll reopen and put in the work to fix them.

PR #95, opened 2017-02-24
[lms] [ForumsTasks] new task search_text for random text searching

pwnage101:
This new task should exercise Elasticsearch, unlike search_topic which
claimed to but actually did not.

@max @robrap @dianakhuang The task ratios are still not determined, I may need to do some more digging to make those accurate.  Otherwise the code is good for review.


loadtests/lms/forums.py
```diff
@@ -222,7 +244,8 @@ def forum_form_discussion(self):
     def search_topic(self):
```
robrap: Would `select_topic` be a better name?

loadtests/lms/forums.py
```diff
@@ -222,7 +244,8 @@ def forum_form_discussion(self):
...
         """
         Load a randomly-selected topic from the discussion tab sidebar.
-        Uses ES, but not actually a search from the user perspective.
+
+        This does *not* use Elasticsearch.
         """
         self.get(
             'discussion/forum/search',
```
robrap: I _think_ that we could drop `search` from this url and it will do the same thing and be more intuitive.  I'm not 100% sure, but seems so from my review of the code with @dianakhuang.

loadtests/lms/forums.py
```diff
@@ -233,6 +256,21 @@ def search_topic(self):
             name='forums:search_topic',
```
robrap: Again, change name to match function rename.

loadtests/lms/forums.py
```diff
@@ -13,15 +13,37 @@
...
 
 
-def _dummy_text(minlen, maxlen):
+def _random_word_from_dictionary():
     """
-    Naive helper function to generate dummy text where needed in forums submissions.
+    Get a random word from the dictionary file.
     """
+    global _dictionary_words
+    dictionary_path = os.path.join(os.path.dirname(__file__), DICTIONARY_FILENAME)
```
robrap: Can you add my name to the dictionary? ;)

loadtests/lms/forums.py
pwnage101: That seems like a better name.

loadtests/lms/forums.py
pwnage101: Alright so I confirmed that this code change doesn't result in different behavior.

pwnage101:
According to my observations, select_topic returns zero or more threads which all have the same commentable_id values, whereas search_text returns zero or more threads which may have different commentable_id values.  This makes sense to me, but if you know better, let me know.

loadtests/lms/forums.py
pwnage101: no, you need to be hazed first.

PR #98, opened 2017-03-10
Implement util/merge_settings.py script

pwnage101:
This new script combines multiple settings files given on its arguments.
It will be useful for separating public settings from secrets.  It will
also be useful for passing YAML overrides via jenkins.

---

PR notes:

Nothing currently calls this script, but this unblocks my work on the next ticket about creating a new locust jenkins job DSL, where I'll be able to use this script to do all those things stated above (settings and overrides).

Example usage:

```
$ cat settings_files/example_1.yml
---
SETTING_1: old
SETTING_2: old
---
# secrets below
PASSWORD: # REQUIRED

$ cat settings_files/example_2.yml
---
SETTING_2: new
---
# secrets below
PASSWORD: super-duper-secret

$ util/merge_settings.py settings_files/example_1.yml settings_files/example_2.yml
---
SETTING_1: old
SETTING_2: new
---
PASSWORD: super-duper-secret
```

I did also evaluate a version of this script which preserved dict subkeys, but eventually decided against it because it introduced complexity which I couldn't justify, and was hard to describe/understand the merging process.  That branch still lives: https://github.com/edx/edx-load-tests/tree/pwnage101/perf-392

PERF-392

### DIFF
--- a/helpers/settings.py
+++ b/helpers/settings.py
@@ -34,6 +34,7 @@ tests:
 import os
 import yaml
 import logging
+import copy
 from pkg_resources import resource_filename
 from pprint import pformat
 
@@ -46,7 +47,168 @@ class MissingRequiredSettingError(Exception):
     pass
 
 
-def init(test_module_full_name, required_data=None, required_secrets=None):
+class MalformedSettingFileError(Exception):
+    pass
+
+
+class Settings(object):
+    """
+    Abstraction class for the standard edx-load-tests settings files syntax
+    based on YAML.
+    """
+
+    def __init__(self, data=None, secrets=None):
+        self._data = data or {}
+        self._secrets = secrets or {}
+
+    @property
+    def data(self):
+        return self._data
+
+    @property
+    def secrets(self):
+        return self._secrets
+
+    def __eq__(self, other):
+        return self.data == other.data and self.secrets == other.secrets
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
+    @classmethod
+    def from_file(cls, settings_file):
+        """
+        Factory method to create Settings instances from settings files.
+
+        Arguments:
+            settings_file (file):
+                This open file object represents a settings file.
+
+        Returns:
+            A Settings instance containing the data and secrets from the given
+            stream.
+        """
+        data, secrets = cls.load_file(settings_file)
+        return cls(data, secrets)
+
+    @classmethod
+    def load_file(cls, settings_file):
+        """
+        Load the contents of the open file object as settings.
+
+        Arguments:
+            settings_file (file):
+                This open file object represents a settings file.
+
+        Returns:
+            Two-tuple of dicts, where the first is settings data and the second
+            is settings secrets.
+        """
+        settings_documents = list(yaml.safe_load_all(settings_file))
+
+        # capture the data and secrets from their respective YAML documents
+        data = None
+        secrets = None
+        if len(settings_documents) == 1:
+            data, = settings_documents
+        elif len(settings_documents) == 2:
+            data, secrets = settings_documents
+        elif len(settings_documents) > 2:
+            raise MalformedSettingFileError("The settings file has more than two documents.")
+
+        # YAML treats empty documents as None, we normalize it to an empty dict
+        if data is None:
+            data = {}
+        if secrets is None:
+            secrets = {}
+
+        # Make sure we're actually returning dicts.  Documents in YAML could
+        # contain things other than mappings, but that wouldn't be valid for
+        # settings.
+        # not_mappings is a list of error messages:
+        not_mappings = [
+            "{} has type '{}'".format(descriptor, type(obj).__name__)
+            for descriptor, obj
+            in {"first document": data, "second document": secrets}.items()
+            if not isinstance(obj, dict)
+        ]
+        if not_mappings:
+            # Say "mapping" in the error message instead of "dict" because
+            # that's what they're called in YAML.
+            raise MalformedSettingFileError(
+                "One or more YAML documents in the settings file was not a mapping: {}.".format(', '.join(not_mappings))
+            )
+
+        return (data, secrets)
+
+    def validate_required(self, required_data=(), required_secrets=()):
+        """
+        Validate the settings keys by making sure the required ones are
+        present.
+
+        Arguments:
+            required_data (iterable of str):
+                dict keys which we will be confirming are in self.data and not
+                mapped to None.
+            required_secrets (iterable of str):
+                dict keys which we will be confirming are in self.secrets and
+                not mapped to None.
+
+        Raises:
+            MissingRequiredSettingError: if there are any missing settings keys
+        """
+        missing_data_keys = [key for key in required_data if self.data.get(key) is None]
+        missing_secret_keys = [key for key in required_secrets if self.secrets.get(key) is None]
+        if missing_data_keys or missing_secret_keys:
+            msgs = []
+            if missing_data_keys:
+                msgs.append('Missing settings: {}.'.format(', '.join(missing_data_keys)))
+            if missing_secret_keys:
+                msgs.append('Missing secret settings: {}.'.format(', '.join(missing_secret_keys)))
+            raise MissingRequiredSettingError(' '.join(msgs))
+
+    def dump(self, stream):
+        """
+        Dump the generated settings file contents to the given stream.
+
+        Arguments:
+            stream (file):
+                An open writable file object to dump settings into.
+        """
+        # only dump the secrets yaml document if it is populated
+        docs_to_dump = [self.data]
+        if self.secrets:
+            docs_to_dump.append(self.secrets)
+
+        yaml.safe_dump_all(
+            docs_to_dump,
+            stream=stream,
+            default_flow_style=False,  # Represent objects using indented blocks
+                                       # rather than inline enclosures.
+            explicit_start=True,  # Begin the first document with '---', per
+                                  # our usual settings file syntax.
+        )
+
+    def update(self, other_settings):
+        """
+        Merge other_settings into this one.
+
+        Existing settings are overridden by those in other_settings, and the
+        rest should remain unchanged.
+
+        Arguments:
+            other_settings (Settings):
+                Another settings instance.
+        """
+        # Only work with a deep copy of other_settings, or else we might
+        # transfer nested object references from other_settings into self,
+        # causing self to contain a mix of shared and unshared values.
+        other_settings_copy = copy.deepcopy(other_settings)
+        self._data.update(other_settings_copy.data)
+        self._secrets.update(other_settings_copy.secrets)
+
+
+def init(test_module_full_name, required_data=(), required_secrets=()):
     """
     This is the primary entrypoint for this module.  In short, it initializes
     the global data dict, finds/loads the settings files, and validates the
@@ -56,9 +218,6 @@ def init(test_module_full_name, required_data=None, required_secrets=None):
     global secrets
     if data is not None:
         raise RuntimeError('helpers.settings has been initialized twice!')
-
-    required_data = required_data or []
-    required_secrets = required_secrets or []
 
     # Find the correct settings file under the "settings_files" directory of
     # this package.  The name of the settings file corresponds to the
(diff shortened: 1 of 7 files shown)


pwnage101:
@ormsbee

util/merge_settings.py
```diff
@@ -0,0 +1,103 @@
...
+        settings_filenames (tuple of unicode strings):
+            Paths to settings files (containing one or two yaml documents)
+            ordered from lowest to highest priority.
+    """
+    existing_settings_files = []
+    for filename in settings_filenames:
+        if os.path.exists(filename):
+            existing_settings_files.append(open(filename, 'r'))
+        else:
+            sys.stderr.write('skipping non-existant file: {}\n'.format(filename))
```
ormsbee: Maybe fail fast here? This feels like the kind of thing that could allow a typo to be lost in the middle of other error log messages and cause confusion when things don't seem to work with the deployed artifact.

util/merge_settings.py
```diff
@@ -0,0 +1,103 @@
...
+
+    Arguments:
+        settings_filenames (tuple of unicode strings):
+            Paths to settings files (containing one or two yaml documents)
+            ordered from lowest to highest priority.
+    """
+    existing_settings_files = []
+    for filename in settings_filenames:
+        if os.path.exists(filename):
+            existing_settings_files.append(open(filename, 'r'))
```
ormsbee: Please don't separate the open and close as you're doing. While it's not really dangerous here, in general try to use the `with open(filename) as ... :` pattern so that it cleans up nicely in case something goes wrong.

util/merge_settings.py
pwnage101: That's totally fair.  I had the settings files hard-coded in the recipe for the make target, and this made it possible to do that.  Will change to fail fast.

util/merge_settings.py
pwnage101: Good catch, I feel like this was a relic of another refactor.  Will use `with open(...` down below instead.

pwnage101:
@ormsbee divergence from your proposal:

* instead of a `merged()` classmethod, I made a `update()` instancemethod.
* instead of `from_yaml()` I simply called it `from_file()` but I think we're on the same page.
* I didn't bother with any readonly (getter-only?) properties, but I could still make that happen since there isn't any code outside of the class which sets those instance variables.  My tendency is to leave it as-is.

Now, running the script requires installing the package, so that it gets access to `helpers.settings`.  I may decide to add `-e .` to requirements/base.txt if I can figure out once and for all if that's dirty or not.

helpers/settings.py
```diff
@@ -46,6 +46,125 @@ class MissingRequiredSettingError(Exception):
...
+
+        Returns:
+            A Settings instance containing the data and secrets from the given
+            stream.
+        """
+        instance = cls()
+        instance.load_file(settings_file)
+        return instance
+
+    def load_file(self, settings_file):
```
ormsbee: As things evolve over time, you might find it easier if you have file parsing code separated from the code that instantiates the object. So something like:

```python
@classmethod
def from_file(cls, settings_file):
    data, secrets = cls.parse_settings(settings_file)
    return cls(data, secrets)
```

Having it happen in the same place isn't confusing now, but it can get tangled if this functionality gets more sophisticated over time because you'd have to evaluate the correctness of the parsing based on its side effects, and it might be the case at some future point that `Settings` will grow to do more interesting things to its inputs.

helpers/settings.py
```diff
@@ -46,6 +46,125 @@ class MissingRequiredSettingError(Exception):
...
+        Arguments:
+            settings_file (file):
+                This open file object represents a settings file.
+
+        Returns:
+            A Settings instance containing the data and secrets from the given
+            stream.
+        """
+        instance = cls()
+        instance.load_file(settings_file)
```
ormsbee: As you might have guessed by my suggestion of read -only properties, I tend to favor immutable objects, which admittedly cuts against the grain of Python a bit. Most of that bias comes from getting bitten when trying to debug programs with really painful cases of shared mutable state, where it was really hard to predict the effect of a change because any of the chain of inputs could mutate in ways you didn't expect on edge cases.

The pattern you're using here is to create an empty instance and then inflate it with values via another public method.  But that means that anyone with a reference to a `Settings` object can invoke `load_file` at any time. The behavior of `Settings` objects can be harder to predict if they can completely change their values in mid-lifecycle by loading an entirely new settings file -- particularly if you have data structures that are derived from and should map to a giving `Settings`.

Unless there is a specific use case that needs that functionality, I think it's best to either:

* Not have a `load_file` method at all (as that works by side-effect) and just have `from_file` return the instance with a parsing helper method; or...
* Change `load_file` into `_load_file` to make it clear that it's an internal helper and should not be called publicly.

I tend to favor the first because it's less likely to encourage mutation going forward.

helpers/settings.py
```diff
@@ -46,6 +46,125 @@ class MissingRequiredSettingError(Exception):
...
+
+        Arguments:
+            settings_file (file):
+                This open file object represents a settings file.
+        """
+        settings_documents = list(yaml.load_all(settings_file))
+        if settings_documents[0]:
+            self.data = settings_documents[0]
+        if len(settings_documents) == 2 and settings_documents[1]:
+            self.secrets = settings_documents[1]
```
ormsbee: Is a completely empty file valid? Should it error?

helpers/settings.py
```diff
@@ -46,6 +46,125 @@ class MissingRequiredSettingError(Exception):
...
+        return self.data == other.data and self.secrets == other.secrets   
+
+    def __neq__(self, other):
+        return not self.__eq__(other)
+
+    def data_loaded(self):
+        return len(self.data) > 0
+
+    def secrets_loaded(self):
+        return len(self.secrets) > 0
```
pwnage101: Ohh, I forgot in this case that an empty dict is treated as false when evaluated as boolean.  You're right, they are equivalent.  My definition of "loaded" is not whether the file was opened, but whether there are any keys present.

helpers/settings.py
```diff
@@ -46,6 +46,125 @@ class MissingRequiredSettingError(Exception):
...
+
+    def __init__(self, data=None, secrets=None):
+        self.data = data or {}
+        self.secrets = secrets or {}
+
+    def __eq__(self, other):
+        return self.data == other.data and self.secrets == other.secrets   
+
+    def __neq__(self, other):
+        return not self.__eq__(other)
```
pwnage101: Yea, it should actually be `__ne__`.  Are you sure I shouldn't define both?

https://docs.python.org/2/reference/datamodel.html#object.__eq__

> when defining `__eq__()`, one should also define `__ne__()` so that the operators will behave as expected.

\*shrug\*

ormsbee:
Huh. Maybe I'm getting it confused with Python 3.

pwnage101:
I'm interested in what you think about the `update` member function since you haven't mentioned it yet.  Like `load_file`, `update` has the side effect of mutating member variables, so I feel like for consistency I should also make `update` side-effect-free and just return a new object (and also rename it to `merge` or something).

helpers/settings.py
pwnage101: Ah, yaml.load_all() will return a zero-length iterator on empty file, so the first `if` should check length before getting index `0` if I want to support empty files.  Supporting empty files was my intention.

helpers/settings.py
```diff
@@ -46,7 +46,158 @@ class MissingRequiredSettingError(Exception):
...
+
+        # capture the data and secrets from their respective YAML documents
+        data = None
+        secrets = None
+        if len(settings_documents) == 1:
+            data, = settings_documents
+        elif len(settings_documents) == 2:
+            data, secrets = settings_documents
+        elif len(settings_documents) > 2:
+            raise MalformedSettingFileError("The settings file has more than one document.")
```
ormsbee: Also, it might help to include the name of the settings file.

helpers/settings.py
```diff
@@ -46,7 +46,158 @@ class MissingRequiredSettingError(Exception):
...
+
+        Existing settings are overridden by those in other_settings, and the
+        rest should remain unchanged.
+
+        Arguments:
+            other_settings (Settings):
+                Another settings instance.
+        """
+        self.data.update(other_settings.data)
+        self.secrets.update(other_settings.secrets)
```
ormsbee: So you asked what I thought of this method and the fact that it mutates the state of settings. I'm okay with it. I realize that's a bit inconsistent, but that's mostly me trying to find a middle ground between safety and awkwardness. The more complex data and data sharing gets, the more I tend to favor immutability, even at the cost of being awkward. So for instance, I once worked on this project where we had columns of data that went through many transform steps (30+), and I tried to skew heavily towards immutability because it would be impossible to debug otherwise.

This data is a lot less connected and complex, so the main thought I had was "what are the most likely ways that people can shoot themselves in the foot with this?". I thought that the `update` method was okay in this case, because it's the expected use pattern that a base settings is written and incrementally updated with overrides. I thought having a `load_file` instance method was dangerous because it basically reuses an instance to load an entirely different settings file, which people won't anticipate or account for in their code.

I think the one last thing that might bite people at some point is that there's no clear copy mechanism. For instance, if someone wanted to make a copy:

```python
# This is the safe way
copied_settings = Settings()
copied_settings.update(original_settings)

# But people might just as commonly do the following, which would have
# unintended side-effects.
copied_settings = Settings(original_settings.data, original_settings.secrets)
```

util/merge_settings.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+    if len(settings_files) == 1:
+        # Only one filename was given on the command line, so we might as well
+        # preserve any comments by copying directly to stdout
+        sys.stdout.write(settings_files[0].read())
+    else:
+        merged_settings = Settings()
+        for settings_file in settings_files:
+            merged_settings.update(Settings.from_file(settings_file))
+        merged_settings.dump(sys.stdout)
```
ormsbee: Very readable. :)

pwnage101: haha, yea, not a very big script anymore

pwnage101:
most comments addressed, but I still want to figure out a good way to provide a copy mechanism like you suggested.

Better error messages for the MalformedSettingFileError exception:

```
$ cat settings_files/lms.yml
---
foo: bar
---
# secrets below
PASSWORD:
---
# a third document should cause error

$ locust -f loadtests/lms
[2017-03-20 10:46:33,850] what/INFO/helpers.settings: using settings file: /home/sankey/tmp/edx/edx-load-tests/settings_files/lms.yml
[2017-03-20 10:46:33,853] what/ERROR/stderr: Traceback (most recent call last):
...
[2017-03-20 10:46:33,857] what/ERROR/stderr: MalformedSettingFileError
[2017-03-20 10:46:33,857] what/ERROR/stderr: :
[2017-03-20 10:46:33,857] what/ERROR/stderr: /home/sankey/tmp/edx/edx-load-tests/settings_files/lms.yml: The settings file has more than one document.
```

```
$ cat settings_files/lms.yml
---
bar
---
# secrets below
PASSWORD:

$ locust -f loadtests/lms
[2017-03-20 11:14:42,810] what/INFO/helpers.settings: using settings file: /home/sankey/tmp/edx/edx-load-tests/settings_files/lms.yml
[2017-03-20 11:14:42,811] what/ERROR/stderr: Traceback (most recent call last):
...
[2017-03-20 11:14:42,815] what/ERROR/stderr: MalformedSettingFileError
[2017-03-20 11:14:42,815] what/ERROR/stderr: :
[2017-03-20 11:14:42,815] what/ERROR/stderr: /home/sankey/tmp/edx/edx-load-tests/settings_files/lms.yml: One or more YAML documents in the settings file was not a mapping: first document has type 'str'.
```

```
$ cat settings_files/lms.yml
---

---
# secrets below
[bar]

$ locust -f loadtests/lms
[2017-03-20 11:15:30,664] what/INFO/helpers.settings: using settings file: /home/sankey/tmp/edx/edx-load-tests/settings_files/lms.yml
[2017-03-20 11:15:30,666] what/ERROR/stderr: Traceback (most recent call last):
...
[2017-03-20 11:15:30,670] what/ERROR/stderr: MalformedSettingFileError
[2017-03-20 11:15:30,670] what/ERROR/stderr: :
[2017-03-20 11:15:30,670] what/ERROR/stderr: /home/sankey/tmp/edx/edx-load-tests/settings_files/lms.yml: One or more YAML documents in the settings file was not a mapping: second document has type 'list'.
```

pwnage101:
latest push changes the update function to use a deep copy of the other_settings object, so that update does not result in a Settings object with mixed shared and unshared references.

pwnage101:
@ormsbee

ormsbee:
I don't see the copy...?

pwnage101:
Oops, actually ran `git add` this time.

pwnage101:
Most recent push is just splitting out one commit into two, and improving the commit messages.

PR #99, opened 2017-03-10
Added loadtests for devstack

clintonb:


clintonb:
@efagin @robrap @feanil @staubina @gsong @jzoldak @cpennington please review and/or recommend additional tests.

clintonb:
@edx/testeng please review

benpatterson:
@pwnage101 could you take a look?

robrap:
@feanil Have you brought your comment to @pwnage101's attention for the general LMS tests, etc.?    Either way, I guess I'm doing that now. :)

pwnage101:
For organizational reasons, I have concerns about putting loadtest helper modules in `util/` which traditionally houses scripts and other modules which aren't imported by any of the loadtests.  Can you move `util/mixins.py` to `helpers/mixins.py`?

clintonb:
@pwnage101 I have moved the mixins. Let me know if you have any other feedback.
@feanil I agree that we should exercise the full logistration flow in general. For these specific tests, I just wanted to ensure Docker-based devstack will sustain our needs for development, which don't necessarily involve frequent logins.

pwnage101:
@clintonb LGTM

PR #101, opened 2017-04-18
Add new course home page to the lms tests.

dianakhuang:
We have simply added a new page.  None of the older pages need to be removed.

I noticed that we also had an older task for a page that no longer exists and is not used for anything, so I have removed it.


dianakhuang:
FYI @pwnage101 on updates to new LMS tests.

pwnage101:
I approve, feel free to merge.  Don't let this block you, but I'm curious how you got to that task ratio of 10?

dianakhuang:
@pwnage101 I picked 10 because this is replacing the existing courseware tab, and the courseware tasks were set to 10 as well.

PR #106, opened 2017-06-09
infrastructure for generating summary reports

pwnage101:
This commit adds the "generate_summary" entry point.  It reads the
standard logfile path and generates a summary report on standard output.

In order to collect enough information, also add some markers to the
logging (via helpers.markers).

This functionality is used by parallel changes to the jenkins job DSL: https://github.com/edx/jenkins-job-dsl/pull/162

An example jenkins test run containing two build artifacts (one logfile, one summary file): https://test-jenkins.testeng.edx.org/job/run-simple-loadtest/44/

### DIFF
--- a/helpers/markers.py
+++ b/helpers/markers.py
@@ -1,0 +1,112 @@
+"""
+This module causes logging of test start/stop and other critical events.
+
+The locust logging format is not necessarily stable, so we use the event hooks
+API to implement our own "stable" logging for later programmatic reference.
+
+Enable this feature in a load test by including the following lines in some
+form, near the begginning of the locustfile:
+
+    from helpers import markers
+    markers.install_event_markers()
+
+As of this writing, the events are:
+
+* locust_start_hatching
+* master_start_hatching
+* quitting
+* hatch_complete
+* edx_heartbeat
+"""
+import re
+import logging
+from datetime import datetime, timedelta
+
+LOG = logging.getLogger(__name__)
+
+# As of this writing, locust prefixes logs using the default datefmt used by
+# the python logging library:
+# https://github.com/locustio/locust/blob/v0.7.5/locust/log.py#L13
+# https://github.com/python/cpython/blob/master/Lib/logging/__init__.py#L492
+LOCUST_TIMESTAMP_FORMAT = '%Y-%m-%d %H:%M:%S,%f'
+
+
+class EventMarker(object):
+    """
+    Simple event marker that logs on every call.
+    """
+    def __init__(self, name):
+        self.name = name
+
+    def _generate_log_message(self):
+        LOG.info('locust event: {}'.format(self.name))
+
+    def __call__(self, *args, **kwargs):
+        self._generate_log_message()
+
+
+class HeartbeatEventMarker(EventMarker):
+    """
+    Event marker which behaves like a heartbeat.
+
+    This marker implements a rate limit on logging to prevent flooding logs.
+    """
+    def __init__(self, name='edx_heartbeat', period=timedelta(seconds=30)):
+        super(HeartbeatEventMarker, self).__init__(name)
+        self.period = period
+        self._last_heartbeat = None
+
+    def __call__(self, *args, **kwargs):
+        if not self._last_heartbeat:
+            self._last_heartbeat = datetime.now()
+        elif self._last_heartbeat + self.period < datetime.now():
+            self._generate_log_message()
+            self._last_heartbeat = datetime.now()
+
+
+def install_event_markers():
+    """
+    Call this function from a locustfile to enable event markers in logging.
+    """
+    # "import locust" within this scope so that this module is importable by
+    # code running in environments which do not have locust installed.
+    import locust
+
+    # install simple event markers
+    locust.events.locust_start_hatching += EventMarker('locust_start_hatching')
+    locust.events.master_start_hatching += EventMarker('master_start_hatching')
+    locust.events.quitting += EventMarker('quitting')
+    locust.events.hatch_complete += EventMarker('hatch_complete')
+
+    # install heartbeat markers which are rate limited
+    heartbeat_handler = HeartbeatEventMarker()
+    locust.events.request_success += heartbeat_handler
+    locust.events.request_failure += heartbeat_handler
+
+
+def parse_logfile_event_marker(line_str):
+    """
+    Parse a logfile line as an event marker.
+
+    Parameters:
+        line_str (str): a line from the locust log for a load test with markers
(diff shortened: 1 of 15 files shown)


helpers/markers.py
```diff
@@ -0,0 +1,81 @@
...
+import logging
+from datetime import datetime, timedelta
+
+LOG = logging.getLogger(__name__)
+
+# As of this writing, locust prefixes logs using the default datefmt used by
+# the python logging library:
+# https://github.com/locustio/locust/blob/v0.7.5/locust/log.py#L13
+# https://github.com/python/cpython/blob/master/Lib/logging/__init__.py#L492
+LOCUST_TIMESTAMP_FORMAT = '%Y-%m-%d %H:%M:%S,%f'
```
ormsbee: But this is configurable, right? Because Locust doesn't specify it, it'll match whatever the logging configuration is?

helpers/markers.py
```diff
@@ -0,0 +1,81 @@
...
+    locust.events.quitting += EventMarkers.event_handler_factory('quitting')
+    locust.events.hatch_complete += EventMarkers.event_handler_factory('hatch_complete')
+
+    heartbeat_handler = EventMarkers.heartbeat_event_handler_factory()
+    locust.events.request_success += heartbeat_handler
+    locust.events.request_failure += heartbeat_handler
+
+
+def parse_logfile_event(line_str):
+    match = re.match(
```
ormsbee: Is this intended to parse files that this PR writes based off of event listeners, or that locust writes by default? Please also specify the file format expectations in the docstring.

util/app_monitors.py
```diff
@@ -0,0 +1,95 @@
+"""
+"""
+try:
+    from urllib.parse import urlencode
+except ImportError:
+    from urllib import urlencode
```
pwnage101: It was an attempt to make transitioning to python 3 easier. http://python-future.org/compatible_idioms.html#urllib-module

helpers/markers.py
pwnage101: Well, it's configurable by locust, but locust opts for the default format.  The default format is not importable from any global module, so I gave up and hard-coded it here.  Better ideas welcome!

util/app_monitors.py
ormsbee: Ah, makes sense. I wonder how big a lift it would be to do the Python 3 conversion at this point (as a separate PR, of course). Nothing really pulls in edx-load-tests as a dependency as far as I know, so we should be able to convert it wholesale to Python 3 and not have to support both simultaneously.

Anyhow, just some random thoughts. This makes total sense as a transition step.

helpers/markers.py
pwnage101: Er, let me actually correct myself, locust specifically uses `%(asctime)s` for the timestamp prefix, and that's essentially hard-coded to `%Y-%m-%d %H:%M:%S,%f` in cpython.  Still, it is not importable, so I hard-code it.

helpers/markers.py
pwnage101: The former.  This function can only read log lines created by the above markers.  Adding docstring.

util/app_monitors.py
pwnage101: Yea, this is tangential but the only thing stopping me from considering a wholesale switchover to python 3 was the fact that the `csm` load test actually imports from lms from edx-platform.  It's unclear to me how much we care about that load test if the `lms` load test does exercise the CSM indirectly.  Perhaps @cpennington you also have some insight regarding the utility of the csm load test?

helpers/markers.py
```diff
@@ -0,0 +1,122 @@
...
+    """
+    Call this function from a locustfile to enable event markers in logging.
+    """
+    # "import locust" within this scope so that this module is importable by
+    # code running in environments which do not have locust installed.
+    import locust
+    locust.events.locust_start_hatching += EventMarkers.event_handler_factory('locust_start_hatching')
+    locust.events.master_start_hatching += EventMarkers.event_handler_factory('master_start_hatching')
+    locust.events.quitting += EventMarkers.event_handler_factory('quitting')
+    locust.events.hatch_complete += EventMarkers.event_handler_factory('hatch_complete')
```
ormsbee: Hey. So I don't think I fully understood what you were doing earlier with the class variable, but I think I get it now. You need the class var to hold the last heartbeat state because Locust expects callable functions? If that's the case, you can use the `__call__` method so that your object is callable and you can treat your object like a function.

```python
class HeartbeatEventMarker(object):
    def __init__(self, name, period):
        self.name = name
        self.period = period
        self._last_heartbeat = None

    def __call__(self):
        if not self._last_heartbeat:
        # etc.
```

And then you'd add it like:

```python
heartbeat_handler = HeartbeatEventMarker(name, period)
locust.events.request_success += heartbeat_handler
locust.events.request_failure += heartbeat_handler
```

So you wouldn't need to create a factory or hold any class global state. Because whether it's concurrency or test cleanup, class members almost always comes back to bite in horrible ways eventually. :-P

helpers/markers.py
```diff
@@ -0,0 +1,122 @@
...
+
+    heartbeat_handler = EventMarkers.heartbeat_event_handler_factory()
+    locust.events.request_success += heartbeat_handler
+    locust.events.request_failure += heartbeat_handler
+
+
+def parse_logfile_event_marker(line_str):
+    """
```
ormsbee: Is this method no longer called?

pwnage101: The caller used a stale name, latest push fixes that.

helpers/markers.py
pwnage101: Cool, I didn't know about `__call__`.  I'll refactor to match your suggestion.

pwnage101:
@ormsbee alright, see latest changes related to the EventMarker callable classes.  The current code was tested here: https://test-jenkins.testeng.edx.org/job/run-simple-loadtest/58/

helpers/markers.py
```diff
@@ -0,0 +1,112 @@
...
+        super(HeartbeatEventMarker, self).__init__(name)
+        self.period = period
+        self._last_heartbeat = None
+
+    def __call__(self, *args, **kwargs):
+        if not self._last_heartbeat:
+            self._last_heartbeat = datetime.now()
+        elif self._last_heartbeat + self.period < datetime.now():
+            self._generate_log_message()
+            self._last_heartbeat += self.period
```
ormsbee: Nit: Isn't the last heartbeat `datetime.now()`? Say there's a long pause, and we suddenly get 10 events all at once?

PR #107, opened 2017-06-19
[teams_discussion] cause clients to retry joining teams

pwnage101:
Before this commit, if a client tries to join a full team, it will fail,
then clog the logs with subsequent failed tasks (mostly create_thread).

After this commit, the client will retry joining different teams until
hopefully one is joined.  Only if a team is successfully joined, the
client will continue running subsequent tasks.

Also update docs to clarify the setup procedure and other
responsibilities.

---

FYI @MichaelRoytman

### DIFF
--- a/loadtests/teams_discussion/locustfile.py
+++ b/loadtests/teams_discussion/locustfile.py
@@ -7,13 +7,21 @@ import sys
 # due to locust sys.path manipulation, we need to re-add the project root.
 sys.path.append(os.path.dirname(os.path.dirname(os.path.dirname(__file__))))
 
+import backoff
 from collections import deque
+import requests.exceptions
 import json
 import random
 import string
 from locust import HttpLocust, task
 from helpers.auto_auth_tasks import AutoAuthTasks
 from helpers import settings, markers
+import logging
+
+# make sure all the retries appear in the log
+logging.getLogger('backoff').setLevel(logging.INFO)
+
+LOG = logging.getLogger(__name__)
 
 settings.init(__name__, required_data=[
     'COURSE_ID',
@@ -24,6 +32,24 @@ settings.init(__name__, required_data=[
 markers.install_event_markers()
 
 _dummy_chars = string.lowercase + ' '
+
+
+class TeamFullException(Exception):
+    """
+    Helper exception for the retry infrastructure for joining teams.
+    """
+    pass
+
+
+def join_team_giveup_handler(details):
+    """
+    This handler is called when a locust client gives up joining a team.
+
+    This giveup handler corresponds to TeamsDiscussionTasks.join_team()
+    retries.  We assume that this point has been reached if all the teams are
+    full.
+    """
+    raise RuntimeError('Check that there are enough open memberships among the teams.')
 
 
 def _dummy_text(minlen, maxlen):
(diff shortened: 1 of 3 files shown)


requirements/base.txt
```diff
@@ -8,6 +8,7 @@ pyzmq==15.2.0  # needed for distributed mode. remove after we start using locust
 # common loadtest requirements
 edx-opaque-keys==0.4
 lazy
+backoff
```
ormsbee: Pin to a version please (or a <= 2.0 if you trust they're doing semantic versioning right).

loadtests/teams_discussion/locustfile.py
```diff
@@ -95,20 +121,38 @@ def _pick_team(self):
...
+                          max_tries=15,
+                          interval=2,  # 15 max tries and 2 second interval corresponds to possibly 30 seconds of
+                                       # retries until this client gives up.
+                          on_giveup=join_team_giveup_handler)
     def join_team(self):
         """
         Get the current user to join a team.
+
+        In case a client tries to join a full team, this function is decorated
+        to retry up to 8 times until an open slot is found (randomly).  As of
```
ormsbee: Does the docstring need to be updated to "15 times" here?

loadtests/teams_discussion/locustfile.py
```diff
@@ -95,20 +121,38 @@ def _pick_team(self):
...
+        this writing, there is no obvious way to determine if a team is full
+        via the teams API.
```
ormsbee: That... is really unfortunate. :-P

loadtests/teams_discussion/locustfile.py
```diff
@@ -26,13 +34,31 @@
...
     """
     Generate dummy text for forum submissions.
     """
     return "".join(random.choice(_dummy_chars) for _ in xrange(minlen, random.randrange(minlen + 1, maxlen)))
 
 
+# This function needs to be defined in the top level scope for this module
+# because the backoff decorator errored when I tried to use it as a
+# TeamsDiscussionTasks classmethod: "name 'TeamsDiscussionTasks' is not
+# defined". -Troy
```
ormsbee: Right, at the time the decorator init code is run, we haven't completed the definition of the class yet, so it can't be referred to. But I believe you can refer to the class method directly like:

```python
class TeamsDiscussionTasks(AutoAuthTasks):
    @classmethod
    def join_team_giveup_handler(cls, details):
        raise RuntimeError('Check that there are enough open memberships among the teams.')

    @backoff.on_exception(on_giveup=join_team_giveup_handler)
    def join_team(self):
        # etc.
```

pwnage101:
Alrighty, commets addressed!  I've never used the backoff package before so I'm not really sure how trustworthy its versioning is.

pwnage101:
oh, actually i'm seeing a new error related to the handler method:

https://build.testeng.edx.org/job/loadtest-driver/134/console

```
TypeError: 'classmethod' object is not callable
```

ormsbee:
Ugh, never mind the classmethod bit then. :-( Sorry for the wasted test cycle.

pwnage101:
Np, it was a nice try.  Giveup handler back to global scope, with improved docstring.

Giving up looks correct: https://build.testeng.edx.org/job/loadtest-driver/135/console

PR #108, opened 2018-01-08
Completion API load test

ssemenova:
A load test to test the completion API by sending completion data for HTML and video xblocks, which are "completed" disproportionately more often than any other xblock type and can represent an upper bound on how many times the completion API is hit.

### DIFF
--- a/loadtests/completion_api/locustfile.py
+++ b/loadtests/completion_api/locustfile.py
@@ -1,0 +1,126 @@
+"""
+Load test for the Course Completion API performance
+"""
+import os
+import sys
+
+sys.path.append(os.path.dirname(os.path.dirname(os.path.dirname(__file__))))
+
+
+from locust import HttpLocust, TaskSet, task, events
+from locust.clients import HttpSession
+
+from helpers import settings, markers
+from helpers.api import LocustEdxRestApiClient
+
+
+# due to locust sys.path manipulation, we need to re-add the project root.
+sys.path.append(os.path.dirname(os.path.dirname(os.path.dirname(__file__))))
+
+import requests.exceptions
+import json
+import random
+import string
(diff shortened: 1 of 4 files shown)


jcdyer:
As the author of much of the completion code, I'd be interested in hearing the results of this when it gets run. Please FYI me on any other related PRs if you don't mind.

ssemenova:
@jcdyer Definitely, I'll let you know when it's done.

loadtests/completion_api/locustfile.py
```diff
@@ -0,0 +1,124 @@
+"""
+Load test to test the performance of the course completion API.
```
schenedx: `Load test for the Course Completion API performance`

loadtests/completion_api/locustfile.py
```diff
@@ -0,0 +1,124 @@
...
+import random
+import string
+from locust import HttpLocust, task
+from helpers.auto_auth_tasks import AutoAuthTasks
+from helpers import settings, markers
+
+settings.init(__name__, required_data=[
+    'COURSE_ID',
+    'LOCUST_MIN_WAIT',
+    'LOCUST_MAX_WAIT',
```
schenedx: Seems like this would also require `html` as well as `video` setting values

settings_files/completion_api.yml.example
```diff
@@ -0,0 +1,28 @@
+---
+# The course to drive the load test against. The course should be structured as follows:
```
schenedx: Can you put some comment warning user that values below needs to match the environment of which the test is executing against? For example, if the tests are executed against a sandbox machine, the tester needs to make sure the course id and block ids below are matching to the courseware ids in the sandbox?

PR #109, opened 2018-04-20
WIP Login through the LMS login endpoint

amangano-edx:
Login through the LMS login endpoint, rather than through auto-auth, so that we can test its performance.


pwnage101:
First thought that comes to mind is that this generates interesting load during ramp-up.  I normally wouldn't consider the ramp-up phase of the load test (i.e. when not all locusts have finished running `on_start()`) as actually part of the load test.

Would you consider instead moving this to a task under CoursewareViewsTask which logs off, then logs back in?  In that case you'd need to somehow preserve the generated email from auto_auth, perhaps via the `self.locust` unique instance.

pwnage101:
You're always welcome to use this PR branch to run your test by specifying the branch name in the build parameters.

PR #110, opened 2018-04-23
WIP Add tests for user_api_views

amangano-edx:
Adds tests for the login_session view, and cleans up logic to allow for locust users to switch between TaskSets after ramp-up.


loadtests/lms/base.py
```diff
@@ -44,6 +44,30 @@ def post(self, *args, **kwargs):
...
+        if response.status_code == 200:
+            self.locust._is_logged_in = False
+
     def on_start(self):
-        self.auto_auth()
-        self.enroll(self.course_id)
+        if not (self.locust._email and self._locust._password):
+            self.auto_auth(params={'no_login': True})
+            self.locust._email = self._email
+            self.locust._password = self._password
```
pwnage101: I think I'd prefer the side-effect of self.auto_auth() to store values directly into self.locust instead of self (i.e. the root TaskSet) since that's where we think it should land anyway.

amangano-edx: Ok, I'll consider making that change. I went with this approach for now since other tests (like the credentials and ecommerce tests) expect to be able to pull the credentials directly from the TaskSet.

pwnage101: Fully understanding that there will be a cascading impact by altering the behavior of helpers/auto_auth_tasks.py:

```
$ grep -Ir 'self\._username' loadtests
loadtests/teams_discussion/locustfile.py:            data={'team_id': self.team['id'], 'username': self._username},
loadtests/credentials/locustfile.py:            'username': self._username,
loadtests/credentials/locustfile.py:            'username': self._username,
loadtests/devstack/locustfile.py:        self.client.get('/u/{}'.format(self._username), name='learner_profile')
loadtests/ecommerce/payment.py:            username=self._username,
```

loadtests/lms/base.py
```diff
@@ -44,6 +44,30 @@ def post(self, *args, **kwargs):
         kwargs['headers'] = headers
         return self._request('post', *args, **kwargs)
 
+    def login(self, email, password):
+        response = self.client.post(
+            '/login_ajax',
```
amangano-edx: This should probably be customizable, as it appears that we support several different login endpoints, which may or may not be available depending on LMS configuration.

